PR #23, opened 2022-07-13
Apidocs

alejandrosantanabonilla:
new documentation


docs/source/conf.py
```diff
@@ -10,9 +10,9 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
+import os
+import sys
+sys.path.insert(0, os.path.abspath('../vayesta/'))
```
obackhouse, line 15: Is this just relative to the Makefile? Maybe better as `os.path.abspath(os.path.join(__file__, "..", "..", "..", "vayesta"))`?

README.rst
```diff
@@ -0,0 +1,59 @@
...
+If this is **true**, then you dont need to do anything else of defining origins. Then, to push changes, you just need to do
+
+.. code-block:: console
+
+    git status  
+    git add --all -f  (sometimes the build directory is seen in the .gitignore)
+    git pull
+    git push
+
+and the changes will be done in the tutorials branch.
```
obackhouse, line 59: Have you misplaced this file? It is in the root directory where there is already a `README.md`.

docs/source/conf.py
alejandrosantanabonilla: This is not  for Makefile, this is where sphinx-apidocs is looking for te module list. This is the way has been suggested in forums:

https://stackoverflow.com/questions/17086273/specifying-a-relative-path-in-sphinx-conf-py-for-collaborative-documentation

README.rst
alejandrosantanabonilla: No, this is just a general how to for internal use. You can remove it from the main branch, but I will strongly encourage to leave it, since it keeps a record of what should be done to create the spinx-apidocs from scratch.

obackhouse: Ok - well it needs to go somewhere in the `docs` directory then, it shouldn't be in the root directory. We can have a subsection in the README.md about how to compile the docs, I can do that now. Plus, there is no need to mention git branches if everything will merge into master eventually - this isn't really user-facing info.

obackhouse: In fact I'll hold off on mentioning the docs in the README.md until we host on gh-pages, that will make it a lot easier

alejandrosantanabonilla: Since the documentation is under construction, I think we can have that README.rst in the apidocs branch. I have it also in my own local branch, but it is easier for somebody that is documenting parts of the code to know the procedure to build from scratch using sphinx-apidocs.

obackhouse: Can you want to combine it with the docs/README.rst?

alejandrosantanabonilla: Docstrings are displaying the description within classes and functons

obackhouse:
This can be kept open in case people want to review the online docs before merging - will need to make one last change to the CI before merging this so it pulls from master.

Majokux:
What's the status of this? Should it be merged?
How are docstrings discovered here, does it not use autodoc?

obackhouse:
lgtm, I'd say merge this after the checks and then open a new PR for expanding the coverage of docstrings

PR #25, opened 2022-07-19
Pr/doc adjustments

Majokux:
Small adjustments to the documentation

@obackhouse @alejandrosantanabonilla 
I added the `docs/source/apidoc` directory for `sphinx-apidoc` files, since I think it's good to separate automatically created files in their own directory. I changed the call in `workflows/ci.yaml` to `sphinx-apidoc -o source/apidoc -t templates ../vayesta/` is this all that needs to be done?

@ghb24 I reverted/changed some minor things of your corrections, in case you want to have a look

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,7 +42,7 @@ jobs:
         run: |
           python -m pip install sphinx sphinx_rtd_theme
           cd docs
-          sphinx-apidoc -o source/ ../vayesta/
+          sphinx-apidoc -o source/apidoc -t templates ../vayesta/
           make html
           cd ..
         if: matrix.documentation
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,4 +53,4 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: docs/build/html
           force_orphan: true
-        if: matrix.documentation
+        if: matrix.documentation && github.ref == 'refs/heads/master'
--- a/docs/source/apidoc/modules.rst
+++ b/docs/source/apidoc/modules.rst
@@ -1,0 +1,22 @@
+API
+===
+
+.. warning::
+   The API documentation is work-in-progress!
+
+Subpackages
+-----------
+
+.. toctree::
+   :maxdepth: 1
+
+   vayesta.core
+   vayesta.dmet
+   vayesta.edmet
+   vayesta.ewf
+   vayesta.lattmod
+   vayesta.libs
+   vayesta.misc
+   vayesta.mpi
+   vayesta.rpa
+   vayesta.solver
--- a/docs/source/apidoc/modules.rst
+++ b/docs/source/apidoc/modules.rst
@@ -4,16 +4,23 @@ API
 .. warning::
    The API documentation is work-in-progress!
 
-Subpackages
------------
+Basic
+-----
+
+.. toctree::
+   :maxdepth: 1
+
+   vayesta.dmet
+   vayesta.edmet
+   vayesta.ewf
+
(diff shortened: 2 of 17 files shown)


alejandrosantanabonilla:
Let's waut for the tests to finish and see if the passed, including creating correctly the Docs. @obackhouse , the github.ref is for running the docs only for the master, right? Maybe, the branches workflow keyword might be useful here?

on:
  push:
    branches:
      - '*'
      - '!master'

This is running all tests but in the master one. If not, then we can just leave the github.ref, might be easier to follow the .yml file.

Majokux:
>I think actually the changes will be deployed directly from this PR once the CI gets past installation and tests so you can double check online before merging. That's a bad idea but probably fine for now because it means we can preview the changes made by PRs.

However, if there are multiple PRs, the latter will overwrite the docs of the former?
Can we have two worksflows, once for the master branch deploying the main documentation, and another one for PRs, deploying to a PR specific link?

obackhouse:
> Let's waut for the tests to finish and see if the passed, including creating correctly the Docs. @obackhouse , the github.ref is for running the docs only for the master, right? Maybe, the branches workflow keyword might be useful here?
> 
> on: push: branches: - '*' - '!master'
> 
> This is running all tests but in the master one. If not, then we can just leave the github.ref, might be easier to follow the .yml file.

not sure what you mean here, but the current `ci.yaml` runs on pushes and PRs to master branch, but we only want the docs to deploy on pushes to master (and not PRs). We could of course have a separate workflow for the docs with a separate schedule but it is nice to have them combined. But that's easily controlled by the change I stated above.

alejandrosantanabonilla:
Indeed, that is the point. So the PRs and pushes for all branches except master are carried out by this command. So, I think this command could be used to fork the PRs and push events using the different branches. we can split the workflow and say, all PRs and push events of all other branches will not build the docs. The PRs and Push events of the docs are only built using master....

obackhouse:
> Indeed, that is the point. So the PRs and pushes for all branches except master are carried out by this command. So, I think this command could be used to fork the PRs and push events using the different branches. we can split the workflow and say, all PRs and push events of all other branches will not build the docs. The PRs and Push events of the docs are only built using master....

yes, that is what my suggested change does

alejandrosantanabonilla:
alright, so i understood correctly what you wanted to do. We can have also the option of the branches, giving us the freedom to decide what to push for the different branches...

Majokux:
>yeah, that's one issue. I don't think you can have two on the same repo but I suppose we could open a new repo for testing the deployment. I think I could then set up a workflow in this main repo that deploys onto the testing repo without having to PR to the testing repo manually - if you think it's worth the effort

I think for the time being we can just prevent PRs from deploying the docs, if thats easier

obackhouse:
> Could you add that change? Thanks

done, please just double check that it all passes before merging

Majokux:
So the onlien docs atm don't seem quite right:
in https://boothgroup.github.io/Vayesta/quickstart/ewf.html the code snippets are missing - in the DMET quickstart they are present.
Are these docs generated from the master branch or this PR? If I build them locally from either branch, the code snippets are present

alejandrosantanabonilla:
The docs should be created from this PR, since it is not merged yet. I think the problem is the relative path we are using for referencing the .py files. In ewf, we have the following path:

/../../../vayesta/example/....

whereas in edmet, we have:

/../../examples/ewf/molecules/01-simple-ccsd.py

I remember I made a comment to be merged in tutorials for changing this relative path. In ay case, it is easy to change, we need to change all the relative paths in ewf to the latter form and the correct file will be called.

alejandrosantanabonilla:
I have changed the relative paths in the main/docs folder. I have commited those changes. This should solve the problem

PR #26, opened 2022-07-26
Support for UEWF density matrix based energy calculations

basilib:


obackhouse:
CI is failing for the docs:

> Handler <function _process_docstring at 0x7f84b0746430> for event 'autodoc-process-docstring' threw an exception (exception: pop from an empty deque)

Can you try pushing something to this PR again and see if it reproduces? Otherwise, it might not like some syntax in one of your docstrings

vayesta/ewf/uewf.py
```diff
@@ -224,3 +224,62 @@ def get_atomic_ssz(self, dm1=None, dm2=None, atoms=None, projection='sao', dm2_w
...
+        Parameters
+        ----------
+        global_dm1 : bool
+            Use 1DM calculated from global amplitutes if True, otherwise use in cluster approximation. Default: True
+        global_dm2 : bool
+            Use 2DM calculated from global amplitutes if True, otherwise use in cluster approximation. Default: False
+
+        Returns
+        -------
+        e_corr : float
```
obackhouse, line 242: Maybe the sphinx doesn't like this entry without a description

basilib: Strange that it only fails for python 3.8.

vayesta/core/ao2mo/helper.py
```diff
@@ -130,12 +240,79 @@ def pack_vvvv(vvvv):
...
+    dm2  : ndarry or (ndarray, ndarray, ndarry)
+        Two body density matrix or tuple of aa, ab, bb blocks for UHF
+    eris : _ChemistERIs
+        PySCF ERIs object
+
+    Returns
+    ----------
+    e2   : float
+        Two body energy contribution
+    """
```
Majokux, line 257: I think it's this docstrings which needs to be corrected, i.e. spaces removed

basilib: Okay, I've fixed that too.

vayesta/ewf/uewf.py
obackhouse: The docs are only built on the 3.8 job in the CI

obackhouse:
Should pass now if you trigger the CI

Majokux:
Tried the fast make_rdm1 routine, definitely seems like the crossover point is quite far out - do you have any experience on this @basilib ?

vayesta/ewf/urdm.py
```diff
@@ -208,11 +211,494 @@ def make_rdm1_ccsd_global_wf(emb, ao_basis=False, with_mf=True, t_as_lambda=Fals
...
+            rxy_occ_a, rxy_occ_b = np.dot(cx_occ_a.T, cy_occ_a), np.dot(cx_occ_b.T, cy_occ_b)
+            rxy_vir_a, rxy_vir_b = np.dot(cx_vir_a.T, cy_vir_a), np.dot(cx_vir_b.T, cy_vir_b)
+
+            if svd_tol is not None:
+                def svd(a,spin):
+                    assert(spin == 'a' or spin == 'b')
+                    if spin == 'a':
+                        nonlocal total_sv_a, kept_sv_a
+                    elif spin == 'b':
+                        nonlocal total_sv_b, kept_sv_b
```
Majokux, line 301: No need to differentiate between `total_sv_a` and `total_sv_b` (same for `kept_sv`) - just add them alltogether and get rid of the spin dependence of `svd`

vayesta/ewf/urdm.py
```diff
@@ -208,11 +211,494 @@ def make_rdm1_ccsd_global_wf(emb, ao_basis=False, with_mf=True, t_as_lambda=Fals
...
+                    rxy_occ_a_norm = (sxy_occ_a[0] if len(sxy_occ_a) > 0 else 0.0)
+                    rxy_occ_b_norm = (sxy_occ_b[0] if len(sxy_occ_b) > 0 else 0.0)
+                    rxy_vir_a_norm = (sxy_vir_a[0] if len(sxy_vir_a) > 0 else 0.0)
+                    rxy_vir_b_norm = (sxy_vir_b[0] if len(sxy_vir_b) > 0 else 0.0)
+                else:
+                    rxy_occ_a_norm = np.linalg.norm(rxy_occ_a, ord=2)
+                    rxy_occ_b_norm = np.linalg.norm(rxy_occ_b, ord=2)
+                    rxy_vir_a_norm = np.linalg.norm(rxy_vir_a, ord=2)
+                    rxy_vir_b_norm = np.linalg.norm(rxy_vir_b, ord=2)
+                if (min(rxy_occ_a_norm, rxy_occ_b_norm, rxy_vir_a_norm, rxy_vir_b_norm) < ovlp_tol):
```
Majokux, line 349: I don't think we want to take the minimum, but the maximum between spin channels.
So `min(max(occ_a, occ_b), max(vir_a, vir_b))`

vayesta/ewf/urdm.py
```diff
@@ -208,11 +211,494 @@ def make_rdm1_ccsd_global_wf(emb, ao_basis=False, with_mf=True, t_as_lambda=Fals
...
+        dm1a[np.diag_indices(nocca)] += 1
+        dm1b[np.diag_indices(noccb)] += 1
+
+    if ao_basis:
+        dm1a = dot(emb.mo_coeff[0], dm1a, emb.mo_coeff[0].T)
+        dm1b = dot(emb.mo_coeff[1], dm1b, emb.mo_coeff[1].T)
+
+    # --- Some information:
+    emb.log.debug("Cluster-pairs: total= %d  kept= %d (%.1f%%)", total_xy, kept_xy, 100*kept_xy/total_xy)
+    emb.log.debug("Singular values (α) total= %d  kept= %d (%.1f%%)", total_sv_a, kept_sv_a, 100*kept_sv_a/total_sv_a)
```
Majokux, line 697: Only ASCII output please

vayesta/ewf/urdm.py
basilib: Nice catch.

vayesta/ewf/urdm.py
basilib: I think UTF-8 has been standard on linux for a while now. Must we support users with computers from the 80s?

vayesta/ewf/urdm.py
basilib: Somewhere around ~400 orbitals.

image

vayesta/ewf/urdm.py
basilib: Somewhere around ~400 orbitals.

image

vayesta/ewf/urdm.py
basilib: The optimised routine is becomes faster somewhere around ~400 orbitals.

image

Majokux:
[c113f05](https://github.com/BoothGroup/Vayesta/pull/26/commits/c113f052c0bd645c85694da7615ec6c47030137c) is not an "update" but a bugfix - I'm surprised this wasn't found in the tests, I thought they would cover this now

Majokux:
@basilib Thanks for coding up the `contract_2dm_eris`, but is it not used currently?

vayesta/ewf/urdm.py
Majokux: This has absolutely nothing to do with the OS not supporting UTF-8, but keyboards not having UTF-8 keys

basilib:
> @basilib Thanks for coding up the `contract_2dm_eris`, but is it not used currently?

It's not currently used but it will be soon. I'm going to update it to also allow contracting with the 2DM blocks returned from `_gamma2_intermediates ` then use it calculate the cumulant energy for both RHF and UHF.

Majokux:
> It's not currently used but it will be soon. I'm going to update it to also allow contracting with the 2DM blocks returned from _gamma2_intermediates then use it calculate the cumulant energy for both RHF and UHF.

OK, that's a good idea, but please also keep the current version, contracting with a full DM2

basilib:
> OK, that's a good idea, but please also keep the current version, contracting with a full DM2

Sure, a boolean argument to `make_dm_cumulant_e2`.

PR #49, opened 2022-10-05
CAS update+RIRPA floating point overflows

cjcscott:
Quick update to CAS fragmentation functionality, with a minor tweak to RIRPA as well.
This adds reporting of the occupied and virtual energy gaps when adding a fragment, and covers a few more edge cases related to these quantities to ensure degenerate orbitals aren't being split in the cas.

While doing this, I updated us to be consistent with `pyscf`'s CAS specification, so we have
```
with emb.cas_fragmentation() as f:
     f.add_cas_fragment(ncas, nelec)
```
rather than swapping the inputs.

Finally, I had another look at occasional floating-point overflow errors in the RIRPA, which @maxnus originally flagged and I thought were mislabelled underflow errors.
Making changes to ensure suppression of those errors didn't solve the problem, so I had a closer look and managed to identify that very occasionally `np.polynomial.laguerre.laggauss` is failing to correctly generate its expected frequency points.

This can result in a negative frequency point, which should be impossible, and leads to the exponential blowing up to the observed overflow errors.
Given generation requires the roots of the corresponding Laguerre polynomial numerical instability makes some sense, but I can't observe it with any consistency, and it seems for a given interpreter behaviour is deterministically stable or unstable.
As such I've added a check when the quadrature grid is generated for negative frequency points, and in their presence the quadrature is regenerated via the alternative implementation `scipy.special.roots_laguerre`.
With any luck the chances of the two separate (and non-identical) implementations both running into the same problem is vanishing and should catch most cases.

If neither returns a purely positive set of points I've regretfully added a `RuntimeError`, just because even though I think these are low-weighted extremal points that seem not to change our results, it is hypothetically undefined behaviour which I, to be honest, don't totally understand.
If we keep running into issues with this I can add a fall-back to a Curtis-Clenshaw quadrature, which should avoid all these problems at the cost of requiring more quadrature points.

### DIFF
--- a/vayesta/core/fragmentation/cas.py
+++ b/vayesta/core/fragmentation/cas.py
@@ -27,15 +27,15 @@ class CAS_Fragmentation(Fragmentation):
             return orbital_indices, orbital_labels
         raise ValueError("A list of integers is required! orbitals= %r" % orbitals)
 
-    def add_cas_fragment(self, nelec, ncas, name=None, degen_tol=1e-10):
+    def add_cas_fragment(self, ncas, nelec, name=None, degen_tol=1e-10):
         """Create a single fragment containing a CAS.
 
         Parameters
         ----------
-        nelec: int
-            Number of electrons within the fragment.
         ncas: int
             Number of spatial orbitals within the fragment.
+        nelec: int
+            Number of electrons within the fragment.
         name: str, optional
             Name for the fragment. If None, a name is automatically generated from the orbital indices. Default: None.
         """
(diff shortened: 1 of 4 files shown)


maxnus:
```
with emb.cas_fragmentation() as f:
     f.add_cas_fragment(nelec, ncas)
 ```
 The changes indicate that it's the other way round - orbitals first, then electrons? Which is the opposite to the convention used in almost everywhere in quantum chemistry to write CAS(electrons,orbitals)?
 
 >As such I've added a check when the quadrature grid is generated for negative frequency points, and in their presence the quadrature is regenerated via the alternative implementation scipy.special.roots_laguerre.
 
 Is there any downside to just always using the SciPy implementation?
 
 Any idea why the tests aren't running?

**Comment by maxnus** [comment]
>I couldn't remember the general convention, but given dependencies following pyscf.mcscf seems reasonable to me (https://pyscf.org/user/mcscf.html)- I could revert though if people feel strongly?

I don't really mind that much which convention to follow...

>I don't have any reason to think either implementation is going to be better, so just falling back to the alternative in the hope that they have different failure conditions seems the safest approach

Basically I am asking does the SciPy version also have this problem?

**Comment by maxnus** [comment]
What kind of degrees are being used here? Both NumPy and SciPy seem to encounter overflow errors starting from degree 186.

**Comment by cjcscott** [comment]
>Basically I am asking does the SciPy version also have this problem?

I haven't managed to get it to do it, but that's no guarantee that it won't. If the issue is the polynomial solution is numerically ill-posed sometimes then I can't see any reasons one should be better/worse.

>What kind of degrees are being used here? Both NumPy and SciPy seem to encounter overflow errors starting from degree 186.

We default to degree 52, which is sufficient for our uses and shouldn't have any issues per the documentation (hence why this is unexpected!).
Since convergence is exponential if degree 186 is required then it'll be because of a vanishing RPA gap leading to slow integrand decay, for which alternative solutions (eg. analytic computation of lowest eigenvalues) would be required.

**Comment by maxnus** [comment]
>We default to degree 52, which is sufficient for our uses and shouldn't have any issues per the documentation (hence why this is unexpected!).
Since convergence is exponential if degree 186 is required then it'll be because of a vanishing RPA gap leading to slow integrand decay, for which alternative solutions (eg. analytic computation of lowest eigenvalues) would be required.

Which is the example where this happened? I tried all degrees from 1 to 300 with NumPy and SciPy 100 times each, and it never gave me a negative frequency - it did complain about overflows starting from degree 186 however (if you are confident that we never need to go beyond 185, we should probably add a `npoints = min(npoints, 185)` or `if npoints > 184: raise ....` somewhere

**Comment by cjcscott** [comment]
Also, test are running again nicely!

**Comment by cjcscott** [comment]
> OK nice, where in this commit did you actually change the behaviour wrt to a? This except (RuntimeError, NIException) ?

That combined with the adding
```
        if a < 0:
            raise NIException("Negative quadrature scaling factor not permitted.")
```
to all semi-infinite quadrature grid generating functions. All our infinite grids are symmetric, so a negative scaling isn't actually a problem there.

I did also consider just optimising using `abs(a)` instead of `a`, but thought that this initial newton search is just if there's a straightforward root to find convergence is unlikely once we hit this edge case.
This is also the reason results weren't affected- because it's a rootfinding routine once we're getting into numerical overflow territory it's not going to find a root, so falls back to approximating via minimising the absolute value.

>Just to double check, we think there is no NumPy/SciPy error for degrees < 100 in the end?

Nah, I ran some tests and everything seems to be stable there- it's the numerical rootfinding optimising `a` that was providing the numerical instability, rather than that in the quadrature generation.

cjcscott:
If everyone's satisfied with this, I can merge to master?

maxnus:
> If everyone's satisfied with this, I can merge to master?
Sure, will do

PR #52, opened 2022-10-06
Fix bug in UHF 2dm cumulant 

basilib:
Contraction of vvVV block with eris when nvira xor nvirb is zero, fails.
Added check to skip that contribution when necessary.


maxnus:
Thanks for reporting! Can you verify that the alternative fix I pushed also works for your case?

basilib:
Yeah, that works too.

I've added the autoconversion of ROHF to UHF to this PR as well.

maxnus:
> I've added the autoconversion of ROHF to UHF to this PR as well.

OK great. Alternatively, the conversion could be done in `qemb.init_mf`, which is called by all subclasses.

basilib:
That was how I tried to do it initially but setting is_rhf, is_uhf and spinsym appropriately didn't work. Maybe there is some other key attribute I missed.

maxnus:
You wouldn't need to set these - anyway this way is fine too

maxnus:
Please use more meaningful branch name than `fix` in future - thanks :+1:

PR #65, opened 2022-10-21
V1.0.2

maxnus:
WIP


maxnus:
I will probably just change most of the tests to use the legacy bath options, rather than update the test value, and add a single test for the projected bath - over time we can update the tests

ghb24:
Sounds good - thanks.

maxnus:
I'm also gonna change the example, since in the Li2 molecule the DMET bath orbital is maximally entangled, such that the projection has no effect  😅

maxnus:
Hi Charlie, I think in principal this can merged, I was just waiting for #69 - however, maybe it's better to merge this now as you said and rebase #69 onto master or a new dev branch...?

obackhouse:
> Hi Charlie, I think in principal this can merged, I was just waiting for #69 - however, maybe it's better to merge this now as you said and rebase #69 onto master or a new dev branch...?

#69 still needs quite some thought to actually get it functional, and then significant refactoring to clean it up once it's working - so I wouldn't wait on it

PR #69, opened 2022-11-03
Added to_cisdtq functionality to RFCI wave function types for extract…

ghb24:
…ion of C1 -> C4 info


vayesta/core/types/wf/ccsdtq.py
```diff
@@ -14,12 +14,13 @@ def CCSDTQ_WaveFunction(mo, *args, **kwargs):
 
 class RCCSDTQ_WaveFunction(wf_types.WaveFunction):
 
-    def __init__(self, mo, t1, t2, t3, t4):
+    def __init__(self, mo, t1, t2, t3, t4_abab, t4_abaa):
```
maxnus, line 17: Can we keep `t4` as it was - if it requires two spin-blocks even for RHF, we can simple store them in a tuple of size 2,
with the "abaa" block first. This will then agree with the future signature of UCCSDTQ, which will have many blocks.

vayesta/core/types/wf/cisdtq.py
```diff
@@ -14,24 +14,181 @@ def CISDTQ_WaveFunction(mo, *args, **kwargs):
 
 class RCISDTQ_WaveFunction(wf_types.WaveFunction):
 
-    def __init__(self, mo, c0, c1, c2, c3, c4):
+    def __init__(self, mo, c0, c1, c2, c3, c4_abab, c4_abaa):
```
maxnus, line 17: Same as above for RCCSDTQ

vayesta/core/util.py
```diff
@@ -107,6 +107,34 @@ def func_with_doc(func):
 
 # --- NumPy
 
+def tril_indices_ndim(n, dims, include_diagonal=False):
```
maxnus, line 110: I would keep this function in the same file, until it is needed by multiple modules - unless you foresee that this will be the case

vayesta/core/types/wf/cisdtq.py
```diff
@@ -14,24 +14,181 @@ def CISDTQ_WaveFunction(mo, *args, **kwargs):
...
+        # As a useful intermediate, compute t2_aa from t2 (_ab), which is 
+        # just the antisymmetric permutation.
+        t2aa = t2 - t2.transpose(1,0,2,3)
+
+        t1t2 = einsum('ia, jkbc -> ijkabc', t1, t2)
+        t1t1t1 = einsum('ia,jb,kc -> ijkabc', t1, t1, t1)
+
+        t3 -= t1t2                                          
+        t3 += t1t2.transpose(0,1,2,5,4,3)
+        t3 += t1t2.transpose(2,1,0,3,4,5)
```
maxnus, line 42: Question, is it possible to deduce the sign based on the permutation - and possible also the list of permutation itself?
If yes, this could prevent some bugs, which would be difficult to find otherwise

vayesta/core/types/wf/ccsdtq.py
ghb24: Yes - good idea. Done.

vayesta/core/types/wf/cisdtq.py
ghb24: Done.

vayesta/core/util.py
ghb24: It is a relatively long, general utility function, and if it is just defined in the types/wf/fci.py file, it would certainly be overlooked for other potential uses. If we want to do more >2nd order excitation methods, it would certainly be of use there. Unless you feel strongly, I would vote to leave it in a util.py file, which certainly feels like the right place.

vayesta/core/types/wf/cisdtq.py
ghb24: Yes to both, and that is certainly the aim longer term when we can procedurally generate these expressions. @obackhouse  will look into this, but for the meantime, we will leave it like this and test correctness in other ways.

vayesta/core/types/wf/ccsdtq.py
```diff
@@ -14,13 +14,15 @@ def CCSDTQ_WaveFunction(mo, *args, **kwargs):
...
         super().__init__(mo)
         self.t1 = t1
         self.t2 = t2
         self.t3 = t3
-        self.t4_abab = t4_abab
-        self.t4_abaa = t4_abaa
+        self.t4 = t4
+        if not (isinstance(t4, tuple) and len(t4) == 2):
+            raise ValueError('''t4 definition in RCCSDTQ wfn requires 
+                    tuple of (abaa, abab) spin signatures''')
```
maxnus, line 25: This message will contain a newline and many spaces, due to the the multiline string.
Note also that I allow myself up to 120 columns, so I wouldn't break here.
As a side note, I try to follow the following convention:
* `"` for text or sentences, i.e. everything that will be read by a human (error messages, logging, docstrings).
* `'` for keys, options, or enums: (i.e. `my_dictionary['c2aa']` or `my_function(solver='CCSD')`.

ghb24: Good to know. Changed.

maxnus (approved):

vayesta/core/types/wf/cisdtq.py
obackhouse: for the record I do already have [a function for this in ebcc](https://github.com/BoothGroup/ebcc/blob/5b1133ce424656e79451b5565cb9381203e23c8e/ebcc/util.py#L160)

ghb24: Defo @obackhouse - we need to have a chat soon about integrating more of the ebcc functionality.

vayesta/solver/coupling.py
```diff
@@ -291,23 +291,50 @@ def _get_delta_t_for_extcorr(fragment, fock):
...
+        #       TO TEST: 
+        #           o Exact cf FCI for four electron systems and full CCSD and FCI fragments
+        #           o Exact cf FCI for four electron systems and full bath CCSD and FCI fragments? Or just one? (only if zero projectors?)
+        #           o Rotate (ov) space and ensure invariant
+        #           o Extensivity checks for separated fragments? (Two LiH clusters separated should be exact, then check approximate systems)
         # --- T1
         # T3 * V
+        # V are antisymmetrized integrals of V[i,j,a,b] = 2<ij|ab> - <ij|ba>
+        # Vint are the <ij|ab>
+        dt1 += einsum('ijab, jiupab -> up', 2.*govov - govov.transpose(0,3,2,1), self.T3)
```
maxnus, line 324: Wherever possible, capitalized identifiers should only be used for classes in Vayesta.
Avoid line continuation with \ and use parenthesis instead.
What is `self` here - this is not a method?

ghb24: It's updated now.

vayesta/core/types/wf/fci.py
```diff
@@ -85,34 +84,217 @@ def as_cisdtq(self, c0=None):
...
-        c2 = c2.reshape(2*nov).transpose(0,2,1,3)
-        c3 = c3.reshape(3*nov).transpose(0,2,4,1,3,5)
-        c4 = c4.reshape(4*nov).transpose(0,2,4,6,1,3,5,7)
+        assert(len(t3addr) == ijk_pairs * abc_pairs)
+
+        c4_abaa = np.zeros((nocc, nocc, nocc, nocc, nvir, nvir, nvir, nvir))
+        c4_comp = np.zeros((ijk_pairs * abc_pairs, nocc * nvir))
+        for t_cnt_a, trip_ind_a in enumerate(t3addr):
+            # Find alpha i,j,k -> a,b,c indices
+            ijk_alpha = int(t_cnt_a / abc_pairs)
```
maxnus, line 225: Better just use floor division `t_cnt_a // abc_pairs`

vayesta/misc/molecules/molecules.py
```diff
@@ -150,6 +150,18 @@ def ring(atom, natom, bond_length=None, radius=None, z=0.0, numbering=None):
         atoms.append([atom_i, np.asarray([r*np.cos(theta), r*np.sin(theta), z])])
     return atoms
 
+def chain(atom, natom, bond_length=None, numbering=None):
```
maxnus, line 153: bond length shouldn't have `None` as default, if it is required

vayesta/solver/coupling.py
```diff
@@ -280,52 +283,181 @@ def _integrals_for_extcorr(fragment, fock):
         govovbb = eris[ob,vb,ob,vb]
```
maxnus, line 283: I think I wrote this, but this probably needs to be `eris[0]`, `eris[1]`, and `eris[2]` - not that UHF is working anyways

vayesta/solver/coupling.py
```diff
@@ -280,52 +283,181 @@ def _integrals_for_extcorr(fragment, fock):
...
         govov = (govovaa, govovab, govovbb)
-    return fov, govov
+        gvvov = None
+        gooov = None
+        govoo = None
+        raise NotImplementedError
+    return fov, govov, gvvov, gooov, govoo
 
+def _get_delta_t_for_extcorr(fragment, fock, solver, include_t3v=True):
+    ''' Make T3 and T4 correction to CCSD wave function for given fragment.
```
maxnus, line 293: Docstrings with `"""` and no space before first letter please

vayesta/solver/coupling.py
```diff
@@ -399,13 +533,21 @@ def externally_correct(solver, external_corrections):
...
+        fov_x, govov_x, gvvov_x, gooov_x, govoo_x = _integrals_for_extcorr(fx, fock)
+
     for y, corrtype, projectors in external_corrections:
 
         fy = frag_dir[y] # Get fragment y object from its index
         assert (y != fx.id)
 
-        if corrtype == 'external':
-            dt1y, dt2y = _get_delta_t_for_extcorr(fy, fock)
+        if corrtype == 'external' or 'external-fciv':
```
maxnus, line 547: This `or` is probably not what you wanted here 😅

vayesta/solver/coupling.py
```diff
@@ -427,16 +569,39 @@ def externally_correct(solver, external_corrections):
...
+            eris = kwargs['eris']
+            mo_energy = eris.mo_energy
+            nocc = t1.shape[0]
+            eia = mo_energy[:nocc,None] - mo_energy[None,nocc:]
+            eijab = mo_energy[:nocc,None,None,None] + mo_energy[None,:nocc,None,None] \
+                    - mo_energy[None,None,nocc:,None] - mo_energy[None,None,None,nocc:]
+
+            # Divide by energy denominators
+            t1[:] += (dt1 / eia)
+            t2[:] += (dt2 / eijab)
```
maxnus, line 604: This breaks the `delta-tailor` approach, or not?

vayesta/core/types/wf/fci.py
ghb24: Done

vayesta/solver/coupling.py
ghb24: Made a comment in code to address this when it comes to implementing UHF

vayesta/solver/coupling.py
ghb24: Done

vayesta/solver/coupling.py
```diff
@@ -280,52 +283,181 @@ def _integrals_for_extcorr(fragment, fock):
...
+            solver.log.info("fov block zero: No T3 * f contribution.")
+        # (Fa) (Taba) contraction
+        dt2 += einsum('me, ijmabe -> ijab', fov, t3)
+        # (Fb) (Tabb) contraction
+        dt2 += einsum('me, jimbae -> ijab', fov, t3)
+        solver.log.info("(T3 * F) -> T2 update norm from fragment {}: {}".format(fragment.id, np.linalg.norm(dt2)))
+
+        # --- T4 * V
+        # (Vaa) (Tabaa) contraction
+        t4v = 0.25 * einsum('mnef, ijmnabef -> ijab', antiphys_g, t4_abaa)
```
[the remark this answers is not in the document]
ghb24: Done

vayesta/solver/coupling.py
ghb24: Indeed - thanks for the catch!

vayesta/solver/coupling.py
ghb24: Yes - fixed in latest push.

examples/ewf/molecules/04-bath-options.py
```diff
@@ -45,6 +45,11 @@
 # of BNOs for a given threshold, and results in a smaller bath for a given threshold.
 # Other options (not shown) include projecting out all couplings to the DMET bath space, or just projecting
 # one index.
+# The options below are default
+emb_mp2_projdmet = vayesta.ewf.EWF(mf, bath_options=dict(bathtype='mp2', threshold=1e-4, \
```
maxnus, line 49: No `\` needed

vayesta/core/qemb/fragment.py
```diff
@@ -331,20 +331,23 @@ def cluster(self, value):
...
-        if reset_eris:
-            self._eris = None
-            self._seris_ov = None
-        self._results = None
+    def reset(self, reset_bath=True, reset_cluster=True, reset_eris=True, reset_inactive=True):
+        self.log.debugv("Resetting %s (reset_bath= %r, reset_cluster= %r, reset_eris= %r, reset_inactive= %r)",
+                self, reset_bath, reset_cluster, reset_eris, reset_inactive)
+        if not reset_inactive and not self.active: 
+            return
+        else:
```
maxnus, line 339: `else` clause unnecessary

vayesta/core/scmf/scmf.py
```diff
@@ -85,7 +85,8 @@ def kernel(self, *args, **kwargs):
             self.log.info("%s==============", len(self.name)*"=")
 
             if self.iteration > 1:
-                self.emb.reset()
+                # Don't reset inactive clusters for external corrections
+                self.emb.reset(reset_inactive=False)
```
maxnus, line 89: I would suggest two ways to combine `scmf` with external corrections:
* add a `callback` to `scmf`, which is called each iteration before the kernel,
and can be used to update the FCI clusters.
* support the FCI fragments more natively, for example introduce an "auxiliary" flag for fragments, with the following effects:
   * Both auxiliary and non-auxiliary fragments are solved, when calling `emb.kernel()`
   * Auxiliary fragments are solved before any non-auxiliary fragments
   * Auxiliary fragments do not contribute to expectation values
   
   This would then get around the whole "setting active/inactive" script level code

vayesta/core/types/wf/cisdtq.py
```diff
@@ -21,17 +21,179 @@ def __init__(self, mo, c0, c1, c2, c3, c4):
         self.c2 = c2
         self.c3 = c3
         self.c4 = c4
+        if not (isinstance(c4, tuple) and len(c4)==2):
```
maxnus, line 24: Spaces around `==`

vayesta/ewf/fragment.py
```diff
@@ -143,21 +146,27 @@ def add_external_corrections(self, fragments, correction_type='tailor', projecto
                 'tailor': replace CCSD T1 and T2 amplitudes with FCI amplitudes.
                 'delta-tailor': Add the difference of FCI and CCSD T1 and T2 amplitudes
                 'external': externally correct CCSD T1 and T2 amplitudes from FCI T3 and T4 amplitudes.
+                'external-fciv': externally correct CCSD T1 and T2 amplitudes from FCI T3 and T4 amplitudes (note T3V term contracted with integrals from the cluster providing the constraints). 'external' is the same as this.
+                'external-ccsdv': externally correct CCSD T1 and T2 amplitudes from FCI T3 and T4 amplitudes (note T3V term contracted with integrals from the cluster being constrained). Should be more accurate?
             Default: 'tailor'.
         projectors: int, optional
             Maximum number of projections applied to the occupied dimensions of the amplitude corrections.
             Default: 1.
+        test_extcorr: bool, optional
```
maxnus, line 155: Is this for debugging or should stay in general?

vayesta/tests/ewf/test_extcorr.py
```diff
@@ -0,0 +1,263 @@
...
+    def test_r_exact_ec_lih_proj2_ccsdv_store(self):
+        return self._test(('lih_ccpvdz', 'rhf'), proj=2, mode='external-ccsdv', store_wf_ccsdtq=True)
+    def test_r_exact_ec_lih_proj0_ccsdv_nostore(self):
+        return self._test(('lih_ccpvdz', 'rhf'), proj=0, mode='external-ccsdv', store_wf_ccsdtq=False)
+    def test_r_exact_ec_lih_proj1_ccsdv_nostore(self):
+        return self._test(('lih_ccpvdz', 'rhf'), proj=1, mode='external-ccsdv', store_wf_ccsdtq=False)
+    def test_r_exact_ec_lih_proj2_ccsdv_nostore(self):
+        return self._test(('lih_ccpvdz', 'rhf'), proj=2, mode='external-ccsdv', store_wf_ccsdtq=False)
+
+    def _test_10_u4_2imp(self, mode, proj):
```
maxnus, line 77: If this shares nothing with the above, better make this a separate class

vayesta/ewf/fragment.py
ghb24: I see no harm in this staying in general. It will compare two different approaches to get the corrections, and ensure they are the same. It could be changed to a unit test, but I think it reasonable to also have in the code more generally, so that if something strange is going on, bugs in the EC implementation can be rules out easily.

vayesta/core/types/wf/ccsdtq.py
```diff
@@ -13,17 +13,27 @@ def CCSDTQ_WaveFunction(mo, *args, **kwargs):
 
 
 class RCCSDTQ_WaveFunction(wf_types.WaveFunction):
+    # TODO: Contract T4's down to intermediates to reduce EC-CC memory overheads.
```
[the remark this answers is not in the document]
ghb24: Well, I guess we would make a new wave function class, which has T1, T2, T3, and some additional intermediates. It wouldn't be part of the ccsdtq class (though I'm not sure what we would call it)...

vayesta/tests/ewf/test_extcorr.py
```diff
@@ -0,0 +1,263 @@
...
+    def test_hub_ec_sym_proj1_ccsdv(self):
+        return self._test_10_u2_eccc_sym(mode='external-ccsdv', proj=1)
+    def test_hub_ec_sym_proj2_ccsdv(self):
+        return self._test_10_u2_eccc_sym(mode='external-ccsdv', proj=2)
+    def test_hub_ec_sym_proj1_dtailor(self):
+        return self._test_10_u2_eccc_sym(mode='delta-tailor', proj=1)
+    def test_hub_ec_sym_proj1_tailor(self):
+        return self._test_10_u2_eccc_sym(mode='tailor', proj=1)
+
+    def _test_water_ec_regression(self, mode=None, projectors=None):
```
maxnus, line 221: separate class again?

vayesta/core/types/wf/ccsdtq.py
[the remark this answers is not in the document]
maxnus: It can be a specialized `ExternalCorrection` class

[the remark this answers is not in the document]
ghb24: Yeah, exactly, but inheriting from `WaveFunction`

vayesta/core/scmf/scmf.py
ghb24: I guess the second option would likely be cleaner (especially for any end-user). I'm trying to think of drawbacks (we are getting quite a few fragment flags now, but perhaps thats ok...). I suggest we go for that one, unless you can think of a reason not to...?

vayesta/core/qemb/qemb.py
```diff
@@ -1542,7 +1542,7 @@ def reset(self, *args, **kwargs):
     def update_mf(self, mo_coeff, mo_energy=None, veff=None):
         """Update underlying mean-field object."""
         # Chech orthonormal MOs
-        if not np.allclose(dot(mo_coeff.T, self.get_ovlp(), mo_coeff) - np.eye(mo_coeff.shape[-1]), 0):
+        if not np.allclose(dot(mo_coeff.T, self.get_ovlp(), mo_coeff) - np.eye(mo_coeff.shape[-1]), 0.):
```
maxnus, line 1545: They should be the same? 0 is probably more pythonic tbh

vayesta/core/scmf/scmf.py
```diff
@@ -100,7 +100,12 @@ def kernel(self, *args, **kwargs):
 
             dm1 = self.mf.make_rdm1()
             # Check symmetry
-            self.emb.check_fragment_symmetry(dm1)
+            try:
+                self.emb.check_fragment_symmetry(dm1)
+            except:
```
maxnus, line 105: Could you please define a `SymmetryError` in `core/util.py`, change the exception raised in `check_fragment_symmetry` to `SymmetryError` and catch only `SymmetryError` exceptions here - thanks

vayesta/core/types/wf/fci.py
```diff
@@ -450,3 +450,12 @@ def as_cisdtq(self, c0=None):
...
         return wf_types.UCISDTQ_WaveFunction(self.mo, c0, c1, c2, c3, c4, projector=self.projector)
+
+    def as_ccsd(self):
+        return self.as_cisd().as_ccsd()
+
+    def as_ccsdtq(self):
+        return self.as_cisdtq().as_ccsdtq()
+
+    def as_fci(self):
+        return self
```
maxnus, line 624: Are these three not inherited from the spin-restricted class?

vayesta/core/types/wf/fci.py
```diff
@@ -419,17 +443,81 @@ def as_cisdtq(self, c0=None):
...
+                c3_aba[i,k,j,b,c,a] = -c3_comp[d_cnta, s_cntb]
+                c3_aba[j,k,i,b,c,a] = c3_comp[d_cnta, s_cntb]
+
+        if len(t2addra) > 0:
+            # This is a better way of doing it, and then expand! Put rest in a test.
+            assert(np.allclose(c3_comp, np.einsum('i,j,ij->ij', \
+                t2signa, t1signb, self.ci[t2addra[:,None], t1addrb])))
+        del c3_comp
+
+
```
maxnus: Could some of this be put into other auxiliary functions? Even if we don't use those auxiliary functions anywhere else,
it would help to keep the WaveFunction classes more readable.

vayesta/core/types/wf/fci.py
```diff
@@ -479,9 +506,42 @@ def as_cisdtq(self, c0=None):
...
+                symmetry="--------",
+        )
+        c4_aabb = c4_aabb.transpose(0, 1, 4, 5, 2, 3, 6, 7)
+        c4_abbb = decompress_axes(
+                "iajjjbbb",
+                c4_abbb_pack,
+                shape=(nocca, nvirb, noccb, noccb, noccb, nvirb, nvirb, nvirb),
+                symmetry="--------",
+        )
+        c4_abbb = c4_abbb.transpose(0, 2, 3, 4, 1, 5, 6, 7)
```
maxnus, line 544: Hi Ollie, can you please try to stick to the existing way of wrapping lines in Vayesta (fill each line unless you go beyond 120), instead of putting one argument per line - thanks

vayesta/core/types/wf/ccsdtq.py
```diff
@@ -35,5 +35,24 @@ def as_ccsd(self):
...
+    def as_ccsdtq(self):
+        return self
+
+    def as_ccsd(self):
+        if self.projector is not None:
+            raise NotImplementedError
+        return wf_types.UCCSD_WaveFunction(self.mo, self.t1, self.t2)
+
+    def as_cisd(self, c0=1.0):
+        return self.as_ccsd().as_cisd()
```
maxnus, line 58: `as_ccsdtq` and `as_cisd` should be put into the RCCSDTQ_WaveFunction class - for `as_ccsd`, do we not have a `CCSD_WaveFunction`  function, which automatically determines R vs U? Then this could also be put into RCCSDTQ

vayesta/core/types/wf/cisdtq.py
```diff
@@ -211,29 +211,700 @@ def __init__(self, mo, c0, c1, c2, c3, c4):
...
+        t4_bbbbbbbb += einsum("id,ja,kc,lb->ijklabcd", t1_bb, t1_bb, t1_bb, t1_bb) * -1.0
+        t4_bbbbbbbb += einsum("id,jb,ka,lc->ijklabcd", t1_bb, t1_bb, t1_bb, t1_bb) * -1.0
+        t4_bbbbbbbb += einsum("id,jb,kc,la->ijklabcd", t1_bb, t1_bb, t1_bb, t1_bb)
+        t4_bbbbbbbb += einsum("id,jc,ka,lb->ijklabcd", t1_bb, t1_bb, t1_bb, t1_bb)
+        t4_bbbbbbbb += einsum("id,jc,kb,la->ijklabcd", t1_bb, t1_bb, t1_bb, t1_bb) * -1.0
+
+        t1 = (t1_aa, t1_bb)
+        t2 = (t2_aaaa, t2_abab, t2_bbbb)
+        t3 = (t3_aaaaaa, t3_abaaba, t3_abbabb, t3_babbab, t3_bbabba, t3_bbbbbb)
+        t4 = (t4_aaaaaaaa, t4_aaabaaab, t4_aabaaaba, t4_abaaabaa, t4_abababab, t4_bbabbbab, t4_bbbabbba, t4_bbbbbbbb)
```
maxnus: please put this into an additional, auxiliary python file

vayesta/solver/coupling.py
```diff
@@ -281,16 +281,12 @@ def _integrals_for_extcorr(fragment, fock):
...
+        # TODO make consistent with RHF return value
         fov = (fova, fovb)
-        govov = (govovaa, govovab, govovbb)
-        gvvov = None
-        gooov = None
-        govoo = None
-        raise NotImplementedError
+        gooov = (eris[0][oa, oa, oa, va], eris[1][oa, oa, ob, vb], eris[2][ob, ob, ob, vb])
+        govov = (eris[0][oa, va, oa, va], eris[1][oa, va, ob, vb], eris[2][ob, vb, ob, vb])
+        govvv = (eris[0][oa, va, va, va], eris[1][oa, va, vb, vb], eris[2][ob, vb, vb, vb])
```
maxnus: no spaces after commas when slicing, please

vayesta/solver/coupling.py
```diff
@@ -448,9 +446,122 @@ def _get_delta_t_for_extcorr(fragment, fock, solver, include_t3v=True, test_extc
...
 
     elif fragment.base.spinsym == 'unrestricted':
-        raise NotImplementedError
-        # TODO
-        pass
+        # Get ERIs and Fock matrix for the given fragment
+        (f_aa_ov, f_bb_ov), \
+                (v_aaaa_ooov, v_aabb_ooov, v_bbbb_ooov), \
+                (v_aaaa_ovov, v_aabb_ovov, v_bbbb_ovov), \
+                (v_aaaa_ovvv, v_aabb_ovvv, v_bbbb_ovvv) = _integrals_for_extcorr(fragment, fock)
```
maxnus: better to use a collection (dictionary?) or custom data structure here as the return value

obackhouse:
Just FYI @maxnus - much of this functionality is still in development and I've paid no mind to formatting or organisation... once everything is working I will clean it up, with the existing style choices in mind

vayesta/solver/coupling.py
```diff
@@ -448,9 +446,122 @@ def _get_delta_t_for_extcorr(fragment, fock, solver, include_t3v=True, test_extc
...
+        dt2_abab += einsum("kc,kcld,ijlabd->ijab", t1_bb, v_bbaa_ovov, t3_abaaba) * 2.0
+        dt2_abab += einsum("jc,kcld,iklabd->ijab", t1_bb, v_bbbb_ovov, t3_abbabb) * -1.0
+        dt2_abab += einsum("jc,kdlc,iklabd->ijab", t1_bb, v_bbbb_ovov, t3_abbabb)
+        dt2_abab += einsum("kb,kcld,ijlacd->ijab", t1_bb, v_bbbb_ovov, t3_abbabb) * -2.0
+        dt2_abab += einsum("kc,kcld,ijlabd->ijab", t1_bb, v_bbbb_ovov, t3_abbabb) * 2.0
+        dt2_abab += einsum("kc,kdlc,ijlabd->ijab", t1_bb, v_bbbb_ovov, t3_abbabb) * -2.0
+
+        dt1 = (dt1_aa, dt1_bb)
+        dt2 = (dt2_aaaa, dt2_abab, dt2_bbbb)
+
```
maxnus, line 582: move this code to a separate file please

maxnus:
> Just FYI @maxnus - much of this functionality is still in development and I've paid no mind to formatting or organisation... once everything is working I will clean it up, with the existing style choices in mind

OK sure, let me know when it's ready

vayesta/core/util.py
obackhouse: I think this is fine where it is.

vayesta/core/types/wf/cisdtq.py
obackhouse: These expressions are now algorithmically generated, so the permutations are determined algorithmically too.

vayesta/core/types/wf/fci.py
obackhouse: These patterns are using `divmod` now.

vayesta/core/types/wf/fci.py
obackhouse: Yep

vayesta/core/types/wf/fci.py
obackhouse: These functions have slightly more simple syntax now - I think they're fine but I can move their outside of the class if you still have strong opinions

vayesta/core/types/wf/fci.py
obackhouse: All fit in 120 now

vayesta/solver/coupling.py
obackhouse: rip pep8

vayesta/solver/coupling.py
obackhouse: I think this would demand a larger change throughout vayesta to use a namespace or a dictionary instead of tuples, like I've done in `ebcc`.

obackhouse:
@abhishekkhedkar09 can you check the UHF `"external-ccsdv"` tests - they are failing within about 1e-5, but I believe that this functionality should be working now - can you check if this is thresholds or it's actually bugged? We also need to cut some tests out before merging, the runtime is still way too long.

@maxnus you are welcome to review this code if you'd like

maxnus:
@obackhouse 
OK thanks, I will take a look in the next few days

abhishekkhedkar09:
> @abhishekkhedkar09 can you check the UHF `"external-ccsdv"` tests - they are failing within about 1e-5, but I believe that this functionality should be working now - can you check if this is thresholds or it's actually bugged? We also need to cut some tests out before merging, the runtime is still way too long.
> 
> @maxnus you are welcome to review this code if you'd like

@obackhouse : I will fix this. It is correct, they should fail as some of them are regression tests that I could only correctly write now as the functionality is working. I will cut down further on the number of 'fast' marked tests.

examples/ewf/molecules/25-externally-correct.py
```diff
@@ -0,0 +1,62 @@
...
+# Setup the external correction from the CCSD fragment.
+# Two main options: correction_type='external-ccsdv' and 'external-fciv'. For the important T3 * V contribution to the
+#   T2 amplitudes, this determines whether the V is expressed in the FCI or CCSD cluster space. The CCSD cluster is
+#   larger, and hence this is likely to be better, as the correction is longer-ranged (though slightly more expensive).
+#   The other option is 'projectors', which should be an integer between 0 and 2 (inclusive).
+#   The larger the number, the more fragment projectors are applied to the correcting T2 contributions, and less
+#   'bath' correlation from the FCI clusters is used as a constraint in the external correction of the CCSD clusters.
+#   For multiple constraining fragments, proj=0 will double-count the correction due to overlapping bath spaces, and 
+#   in this case, only proj=1 will be exact in the limit of enlarging (FCI) bath spaces.
+ccsd_frag.add_external_corrections(fci_frags, correction_type='external-ccsdv', projectors=1)
```
maxnus, line 48: Are 'external-ccsdv' and 'external-fciv' good names for this option? - basically, I am asking if it is conceivable to have an external correction in this spirit with a different low-level method than CCSD or a different high-level method than FCI - in this case, and in order to future proof keyword arguments, it would be better to refer to 'low_level' or 'high_level' (or similar names), rather than to CCSD and FCI directly.

examples/ewf/other/61-ext-corr-hubbard-1D.py
```diff
@@ -0,0 +1,115 @@
...
+fci_frags = []
+with emb.site_fragmentation() as f:
+    # Set up a two-site FCI fragmentation of full system as auxiliary clusters
+    # Ensure the right number of electrons on each fragment space of the FCI calculation.
+    fci_frags.append(f.add_atomic_fragment([0, 1], solver='FCI', bath_options=dict(bathtype='dmet'), store_wf_type='CCSDTQ', nelectron_target=2*nelectron/nsite, auxiliary=True))
+    # Add single 'complete' CCSD fragment covering all sites
+    ccsd_frag = f.add_full_system(solver='CCSD', bath_options=dict(bathtype='full'), solver_options=dict(solve_lambda=False, init_guess='CISD'))
+for fci_sym_frag in fci_frags[0].add_tsymmetric_fragments(tvecs=[5, 1, 1]):
+    # Add symmetry-derived FCI fragments to avoid multiple calculations
+    fci_frags.append(fci_sym_frag)
```
maxnus, line 42: `fci_frags.extend(fci_frags[0].add_tsymmetric_fragments(tvecs=[5, 1, 1]))`

vayesta/core/types/wf/t_to_c.py
maxnus, line 1: I assume these were auto generated - Could we add a (small) comment to this module how this was done and/or which code was used?

vayesta/solver/coupling.py
```diff
@@ -346,6 +514,9 @@ def _get_delta_t_for_delta_tailor(fragment, fock):
             mo_energy = (moa, mob)
 
     eris = fragment._eris
+    if eris is None:
+        # Symmetry-derived fragments may not have eris stored
+        eris = fragment.base.get_eris_array(cluster.c_active)
```
maxnus, line 519: Does this mean we perform integral transformation for each O(N_k) fragment?
That's OK as an initial solution to explore the method, but we should leave a TODO or similar,
because this wouldn't scale well

maxnus: I just realized this is just for the delta tailoring, which isn't really that relevant (?)

vayesta/solver/coupling.py
```diff
@@ -354,22 +525,28 @@ def _get_delta_t_for_delta_tailor(fragment, fock):
...
-    external_corrections: list[tuple(int, str, int)]
+    external_corrections : list of tuple of (int, str, int)
         List of external corrections. Each tuple contains the fragment ID, type of correction,
         and number of projectors for the given external correction.
+    eris : _ChemistsERIs
+        ERIs for parent CCSD fragment. Used for MO energies in residual contraction, and for
+        type of correction == 'external-ccsdv', where the parent Coulomb integral is contracted.
+        If not passed in, MO energy if needed will be constructed from the diagonal of 
+        get_fock() of embedding base class, and the eris will be also be obtained from the
+        embedding base class. Optional.
```
maxnus, line 545: Why do we not always just use the embedding class attributes?

vayesta/solver/coupling.py
```diff
@@ -383,7 +560,51 @@ def externally_correct(solver, external_corrections):
     cx_vir = cluster.c_active_vir       # Virtual  active orbitals of current cluster
     cxs_occ = spinalg.dot(spinalg.T(cx_occ), ovlp)
     cxs_vir = spinalg.dot(spinalg.T(cx_vir), ovlp)
-
+    # CCSD uses exxdiv-uncorrected Fock matrix for residuals
+    fock = emb.get_fock(with_exxdiv=False)
```
maxnus, line 564: Move fock, to where it is actually being used/needed

vayesta/solver/coupling.py
```diff
@@ -420,17 +640,37 @@ def externally_correct(solver, external_corrections):
...
+        
+        if corrtype in ['external', 'external-fciv', 'external-ccsdv']:
+            # Contract with fragment x (CCSD) energy denominators
+            # Note that this will not work correctly if a level shift used
+            eia = mo_energy[:nocc, None] - mo_energy[None, nocc:]
+            eijab = pyscf.lib.direct_sum('ia,jb->ijab', eia, eia)
+            dt1 /= eia
+            dt2 /= eijab
+
+        solver.log.info("Total external correction amplitudes from all fragments:  dT1= %.3e  dT2= %.3e", \
```
maxnus, line 672: continuation character not needed here

maxnus:
@obackhouse I added a few comments, but looks good (and like a lot of work!) well done

vayesta/solver/ccsdtq.py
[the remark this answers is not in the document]
obackhouse: It's exactly the T3 and T4 residual update in a CCSDTQ solver. So whilst it's not a CCSDTQ solver, it's a part of one. Any suggestions on what to call the file instead?

vayesta/solver/coupling.py
obackhouse: Not sure what you mean here

vayesta/core/types/wf/t_to_c.py
obackhouse: The spin integration tools I have are still closed source for now, I added a note and can update when I open source those tools

obackhouse:
oh and the changelog can just go in the release tag?

cjcscott:
I've also just noticed my conflict resolution commit has triggered possibly the first CI test run on this branch, since they didn't run after your fix for some reason, so we might have some fixes to make there as well!

cjcscott:
>oh and the changelog can just go in the release tag?

As in we don't need to do it until actually the commit where we release v1.0.2? Yeah, I just meant we shouldn't forget to do that if we're sticking to version-specific changelogs. I'm assuming release would correspond to merging #65 into master?

maxnus:
> I'm still waiting on @maxnus to respond to a couple of comments above (though they're minor), and there are unresolved comments about the naming convention and the integral transformation that I think @ghb24 was going to look at. After that, LGTM

I'm happy for this to be merged!

examples/ewf/molecules/25-externally-correct.py
ghb24: In the latest commit, I've removed the different options of 'external-ccsdv' and 'external-fciv', so there is just the 'external' option as an external correction. The choice of 'ccsdv' and 'fciv' is controlled by an additional optional argument to the add_external_corrections function, `low_level_coul`, where True is equivalent to 'external-ccsdv' and False 'external-fciv'. Should be clearer. Tests and examples updated.

vayesta/solver/ccsdtq.py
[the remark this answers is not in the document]
ghb24: I think this is fine. It fits the functionality of a solver, even if it isn't a standalone solver (but one day may be).

vayesta/solver/coupling.py
ghb24: TODO added.

vayesta/solver/coupling.py
ghb24: So the eri's themselves are potentially used here too, when we are contracting with the low-level coulomb integrals. I think for solids, there may be differences in MO energies regarding whether they are the shifted ones or not?

ghb24:
Lets merge this, once we are sure tests pass.

PR #72, opened 2022-12-14
Update rotational symmetry for pyscf compatitbility

basilib:


ghb24 (dismissed):
Thanks Basil!

maxnus:
Thanks for addressing this!
Could you use a `try`-`except` clause or something like this
```
if getattr(pyscf.symm.basis, '_momentum_rotation_matrices'):
    rotfunc = pyscf.symm.basis._momentum_rotation_matrices
else:
    rotfunc = pyscf.symm.basis._ao_rotation_matrices
```
to make sure that the code still works with older versions? Thanks

PR #76, opened 2023-01-27
Fix MPI communication for ERIs larger than 2^31 bytes

basilib:


maxnus:
Hi Basil,
can you comment on the reason for this - can Bcast deal with arrays > 2^31 bytes, whereas bcast cannot?
Is there any downside to just always using Bcast?

basilib:
Bcast will copy directly from a buffer on the root node to the receivers, whereas bcast will make a copy on the root node, serialize the object and then send to receivers. Bcast also does not have the 2^31 bytes (2GB) size limit. We should be using Bcast for all numpy array communication as it is more efficient.

There is a possible complication which may arise if hdf5 files are used in place of numpy arrays, but these should not be transferred via MPI anyway because they are available on the filesystem.

maxnus:
Thanks for clarifying. I added a generalized  `bcast` function to the `MPI_Interface` class, so that the buffer version is used automatically if an array is being broadcasted.

PR #81, opened 2023-03-27
Run tests for PRs to all branches.

cjcscott:
Updates `ci.yaml` to triggers tests to run for all PRs, not just for those to `master` or `dev`.
If this doesn't work to target all PRs we could explicitly filter for all branches within `pull_request`, but that seems unnecessary if this removes filtering entirely.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [master, dev]
   pull_request:
-    branches: [master, dev]
   schedule:
     - cron: '0 2 * * *'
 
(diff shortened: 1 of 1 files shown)


obackhouse:
One side effect of this PR is that @cjcscott will now receive email notifs should the nightly tests fail - @ghb24 if you want these instead, you can push an empty line to the end of the `ci.yaml`?

Again there is probably some more rigorous way to redirect those emails but there's no trivial solution...

ghb24:
> One side effect of this PR is that @cjcscott will now receive email notifs should the nightly tests fail - @ghb24 if you want these instead, you can push an empty line to the end of the `ci.yaml`?
> 
> Again there is probably some more rigorous way to redirect those emails but there's no trivial solution...

Done

Issue #83, opened 2023-04-12
Test failures UHF AFM states


cjcscott:
Hello! This is an issue for discussion of the test failures in #69; I didn't want to confuse the discussion there so thought a different place to discuss would be useful. I was going to add a PR for discussion and a fix, but I'm currently unsure of the best way to nicely fix this.

The test failures are interesting; it's the test from the spin-symmetry-broken UHF fix I merged a while back but it didn't come up till now due to the tests not running for all PRs at that point. I can't reproduce it on my local or zombie python installs (aka all tests pass for me), both of which used the setup.py for all dependencies and seem to be within what we support, but @abhishekkhedkar09 has successfully reproduced it.

It looks to be an error coming from within the pyscf UCCSD ao2mo functionality when called within a UCISD calculation to generate an initial UFCI guess. The explicit error from the logs is
```
>           Lvv[blk] = lib.pack_tril(Lpq[:,va,va].reshape(-1,nvira,nvira))
E           ValueError: cannot reshape array of size 0 into shape (0,0)

../../../.local/lib/python3.7/site-packages/pyscf/cc/uccsd.py:1004: ValueError
```
and the relevant end of the stack trace is
```
vayesta/solver/fci.py:256: in get_cisd_init_guess
    return super().get_cisd_init_guess()
vayesta/solver/fci.py:102: in get_cisd_init_guess
    cisd.kernel()
vayesta/solver/cisd.py:22: in kernel
    if eris is None: eris = self.get_eris()
vayesta/solver/ccsd.py:91: in get_eris
    self.eris = self.base.get_eris_object(self.solver)
vayesta/core/util.py:439: in wrapped
    res = func(self, *args, **kwargs)
vayesta/core/qemb/uqemb.py:196: in get_eris_object
    eris = postscf_ao2mo(postscf, fock=fock, mo_energy=mo_energy, e_hf=e_hf)
vayesta/core/ao2mo/postscf_ao2mo.py:69: in postscf_ao2mo
    eris = postscf.ao2mo(mo_coeff)
../../../.local/lib/python3.7/site-packages/pyscf/ci/ucisd.py:947: in ao2mo
    return uccsd._make_df_eris_outcore(self, mo_coeff)
```
This looks to arise when inferring one shape dimension when another is specified to be zero size for an array of zero size, which makes sense since the obvious approach to calculate this would result in zero divided by zero and that index of the shape doesn't actually change the size of resulting matrix. As a minimal example consider the following:
```
[...]

PR #84, opened 2023-04-13
Remove initial guess for FCI test on dissociated, spin-symmetry-broken H2 test

cjcscott:
This fixes the test failures in #69 , as discussed in #83.
The underlying issues with pyscf and numpy are a bit gnarly, and in the interest of not holding up the original branch I've just removed the calculation of the CISD initial guess which is actually causing the errors and isn't what this actually wants to test. Catching all errors like this is going to be simplified if we have a more standardised approach to calculate our embedded integrals, such as in #66. An alternative tack could be forcing `incore_anyway` in these systems, which might be an OK halfway point.

### DIFF
--- a/vayesta/tests/ewf/test_h2.py
+++ b/vayesta/tests/ewf/test_h2.py
@@ -179,7 +179,10 @@ class Test_UFCI_dissoc(TestCase):
     @classmethod
     @cache
     def emb(cls, bno_threshold):
-        emb = vayesta.ewf.EWF(cls.mf, bath_options=dict(bathtype="dmet"), solver='FCI')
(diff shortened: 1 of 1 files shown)


cjcscott:
I've now (hopefully) implemented a fix to detect this exception, and instead generate the ERIs with `incore_anyway=True` (debugging without being able to reproduce locally is a pain).

In the process of this I've discovered some possible unintentional behaviour in our FCI solver that thwarted my initial approach.  It's not a major problem but I've implemented a quick fix which I'll put in a separate PR to avoid more confusion (using "none" or "None" instead).

PR #85, opened 2023-04-13
Fix fci init_guess

cjcscott:
This fixes the control of initial guesses within the FCI solver, which I came across in the process of trying to fix #84.
I would suggest we merge this after #69 for simplicity; it's not anything high priority.
8502952ee77304ba9141ead8369b36af03d497c9 added the function
```
    def get_init_guess(self):
        if self.opts.init_guess is None:
            return dict(ci0=None)
        if self.opts.init_guess == 'CISD':
            self.log.info("Generating intitial guess from CISD.")
            return dict(ci0=self.get_cisd_init_guess())
        raise ValueError
```
to control the generation of initial guesses for FCI wavefunctions. This seems to have intended passing `solver_opts={"init_guess":None}` to result in an initial guess of the HF determinant, rather than the CISD wavefunction. However, due to inheritance within the code interpreting `None` as the user not passing anything in this actually results in `init_guess="default"(="CISD")` by the time it gets into this function.

As a fix I've just changed this option to instead require "None" or "none". I've also added a test checking the energy of a prematurely terminated calculation correctly starting from the HF initial guess, which is hopefully OK so long as the first step of a davidson is deterministic (which I wouldn't be certain of, but seems OK in initial testing and we'll find out from the CI...).

### DIFF
--- a/vayesta/solver/fci.py
+++ b/vayesta/solver/fci.py
@@ -93,7 +93,7 @@ class FCI_Solver(ClusterSolver):
         if self.opts.init_guess in ["none", "None"]:
             return dict(ci0=None)
         if self.opts.init_guess == 'CISD':
-            self.log.info("Generating intitial guess from CISD.")
+            self.log.info("Generating initial guess from CISD.")
             return dict(ci0=self.get_cisd_init_guess())
         raise ValueError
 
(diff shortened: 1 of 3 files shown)


maxnus:
Hi Charlie, could we call this something more meaningful than `"None"`? Maybe we can try to understand what `fci.get_init_guess()` generates in PySCF?

cjcscott:
The pyscf initial guess for `nroots` takes the states corresponding to the `nroots` lowest energies within `hdiag` as the initial guesses, so for the *n*th root the *n*th lowest energy slater determinant is the initial guess. I've updated this option to instead be `meanfield` or `mf` correspondingly.

PR #88, opened 2023-04-14
Fix dissociated H2 test.

cjcscott:
This fixes the test failures on #69, as discussed in #83, by changing the test to use dense ERIs rather than density fitting. Breaking spin symmetry here required a more explicitly symmetry-broken initial guess than with density fitting, so I've added the suggested approach to break symmetries from `pyscf/examples/scf/32-break_spin_symm.py` into `TestMolecule.uhf_stable`.

This doesn't fix using density fitting in this system- that's proved more difficult (see #84, which I'm going to close now), but given it's an edge case this seems acceptable for now. I would say we could just add a warning/error if a cluster has no excitations in a spin channel and is using density fitting, but this is going to come up again if we're running large Hubbard models.

For now this passes the tests in the only place I can currently produce the original error (thanks Abhi!) so hopefully removes a blocker to #69 merging.

### DIFF
--- a/vayesta/tests/ewf/test_h2.py
+++ b/vayesta/tests/ewf/test_h2.py
@@ -167,8 +167,9 @@ class Test_UFCI_dissoc(TestCase):
 
     @classmethod
     def setUpClass(cls):
-        cls.mf = testsystems.h2_sto3g_dissoc_df.uhf_stable()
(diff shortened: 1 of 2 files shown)


cjcscott:
(I should also note I've also added a TODO in the test referencing the issues with density fitting)

PR #92, opened 2023-05-24
Solver update

cjcscott:
This refactors all solvers to go through a common interface (the `ClusterHamiltonian` classes) for all calculation of integrals and properties. This allows efficient code reuse and (hopefully) a general reduction in complexity when implementing a new solver or changing existing functionality. In general, solver calculations only refer to quantities within the active orbital basis.

Most differently, dummy `pyscf.scf` objects can be generated which full represent an embedded calculation via [customizing the hamiltonian](https://github.com/pyscf/pyscf/blob/8eea7e8cab884bddb7a85aa8ba625503d3a8c9cd/examples/scf/40-customizing_hamiltonian.py); this supports both 3c bare ERIs and 4c bare or screened ERIs. Support for unrestricted calculations with differing number of active orbitals in different spin channels then depends upon freezing orbitals within the solver.

The bones of this are as such the same as discussed in #66, but with some updates focusing on ensuring that integrals can be efficiently accessed when evaluating various expectation values in MP2 and similar calculations. While this is now much more straightforward, there is definitely room for plenty of optimisation in integral storage and caching/conversion- the benefit of this refactor is that any such optimisations will apply to all supported approaches relatively straightforwardly.

I've ensured that all existing functionality is reproduced, though within `solver/coupling.py` I just made sure that all objects are correctly referenced in the new code rather than anything more elaborate. In this part of the code I've also separated the TCCSD and externally corrected CC solvers into their own solvers which inherit from the base CCSD solvers (`TCCSD` and `extCCSD` respectively), as this avoids issues of having lots of incompatible low-level solver options. However, adding external corrections automatically updates the fragment solver to be `extCCSD` if it was previously `CCSD` so practical usage shouldn't be affected.

I've generally tried to add comments and docstrings as I've gone along, but will almost certainly have missed some- please shout if anything is particularly confusing and undocumented!

This leads on to various further updates to the solvers using shared functionality, with the two most imminent being:
- Support for UFCI calculations with different numbers of active orbitals within the cluster (see branch `UFCI_diff_norbs` and #93)
- Support for cRPA screening of cluster interactions, and a rewrite of mRPA and screening functionality in general to be a bit neater (on branch `crpa_screening`)

While there's been some discussion around possibly updating the way `solver_options` are handled, I decided that this PR already contained more than enough changes so have left that be for now.


vayesta/core/qemb/fragment.py
```diff
@@ -1025,3 +1032,30 @@ def plot3d(self, filename, gridsize=(100, 100, 100), **kwargs):
...
+            cluster_solver.hamil.add_screening(self._seris_ov)
+        return cluster_solver
+
+    def get_frag_hamil(self):
+        # This detects based on fragment what kind of Hamiltonian is appropriate (restricted and/or EB).
+        return ClusterHamiltonian(self, self.mf, self.log, screening=self.opts.screening,
+                                  cache_eris=self.opts.store_eris)
+
+    def get_solver_options(self, *args, **kwargs):
+        raise AbstractMethodError
```
obackhouse, line 1061: Add newline character

obackhouse:
Please avoid relative imports - I am going to remove these up for the rest of Vayesta soon but I intend that PR to be merged after this one

cjcscott:
Good point- I'll do a pass removing any I've added

abhishekkhedkar09 (approved):

Issue #91, opened 2023-05-22
SCMF test failures


cjcscott:
I've been getting some weird behaviour in the SCMF tests on master. I get consistent failures which obviously aren't replicated on the CI- specifically, when running `test_scmf.py` I get
```
============================================================================================= short test summary info =============================================================================================
FAILED vayesta/tests/core/scmf/test_scmf.py::SCMF_Test::test_brueckner - AttributeError: 'EWF' object has no attribute 'diis'
FAILED vayesta/tests/core/scmf/test_scmf.py::SCMF_UHF_Test::test_brueckner - AttributeError: 'UEWF' object has no attribute 'diis'
===================================================================================== 2 failed, 2 passed, 3 warnings in 3.36s =====================================================================================
```
These trace back to the SCMF kernel, l79 of `vayesta/core/scmf/scmf.py`
```
        diis = (self.get_diis() if self.diis else None)
```

What seems to be happening here is that since the SCMF kernel is called via the command (in `qemb.py`, with self as the `Embedding` class)
```
        self.kernel = self.with_scmf.kernel.__get__(self)
```
which seems to result in calling `with_scmf.kernel` but with self equal to the embedding object (I'm not hugely familiar with the expected behaviour here, maybe @maxnus knows?).
I'm basing this off the result of running a calculation on the branch `debug_scmf`, which adds logging at the start of `SCMF.kernel`
```
        self.log.warning("In SCMF, self is: %s", str(type(self)))
        self.log.warning(f"self.with_scmf.diis: {self.with_scmf.diis}")
        self.log.warning(f"self.diis: {self.diis}")
```
and gives me the result when running any SCMF calculation of
```
[WARNING] |  In SCMF, self is: <class 'vayesta.ewf.ewf.EWF'>
[WARNING] |  self.with_scmf.diis: True
EWF(mf: DensityFitting-RHF object of <class 'pyscf.df.df_jk.density_fit.<locals>.DensityFitting'>, solver: 'CCSD') <vayesta.core.scmf.pdmet.PDMET_RHF object at 0x7f710d6c9310>
True
Traceback (most recent call last):
  File "/home/charlie/calc/bugfix/scmf/test.py", line 24, in <module>
    rewf.kernel()
  File "/home/charlie/code/Vayesta/vayesta/core/scmf/scmf.py", line 81, in kernel
    self.log.warning(f"self.diis: {self.diis}")
                                   ^^^^^^^^^
[...]

Issue #68, opened 2022-11-03
Example 24 tailoring crashes


ghb24:
I don't know whether this is an arm64 issue, but the example ewf/molecules/24-tailor-with-fragments.py is now crashing as it stops converging the CCSD successfully, and eventually hits a numpy.linalg.LinAlgError: Singular matrix in the DIIS - I don't think it was doing this previously. Can other people see this behaviour?
[observed on both master and v1.0.2a branches]

basilib:
This also occurs on an Intel Xeon cpu with openblas.

PR #95, opened 2023-05-25
Deprecation and cleanup

obackhouse:
I'd like to open this as a discussion and a collaborative effort to clean up code that is no longer in use or has been deprecated for a long time.

- [x] Remove long-deprecated code
- [x] Check that examples are all running non-deprecated code

Examples which are still not running:
edmet
- [x] 70-h-ring-benchmark.py (fermion-boson coupling symmetry broken)
- [x] 71-h-ring-benchmark-df.py (fermion-boson coupling symmetry broken)
- [x] 72-h-chain-benchmark.py (fermion-boson coupling symmetry broken)

ewf/molecules
- [x] 24-delta-tailor.py (convergence)
- [x] 24-tailor-with-fragments.py  (convergence)
- [ ] 71-modify-hcore.py (population analysis needs updating)

ewf/other
- [x] 72-sc-h-plane.py (convergence)
- [x] 76-sc-graphene.py (convergence)


obackhouse:
@maxnus same question as above, but for `vayesta/core/DEL`?

maxnus:
This is a good idea! I made a PR.

I kept some of these files and commented code, since you quickly start to forget previous work, once it's just part of the git history and not that visible anymore. Really these should probably be stale branches.

PMO indeed stands for projected MO, a la regional embedding from Tim Berkelbach. I compared their performance to MP2 BNOs in the past, but the BNOs were much better in the tested systems.

cjcscott:
Just as a follow-on from the SCMF stuff, the functionality https://github.com/BoothGroup/Vayesta/blob/be227313c17c10d183e3dc9c7a522099cc426b6a/vayesta/core/qemb/qemb.py#L228-L235 in `Embedding.__init__`  doesn't seem to be used anywhere, let alone tested, and as demonstrated in #91 uses dodgy syntax. It seems ripe for depreciation unless we have any good reasons not to.

obackhouse:
I've removed `core/DEL` and `core/ao2mo/DEL` - let me know if you want me to restore these into a stale branch. Otherwise at least this PR serves as a checkpoint for removed functionality so they won't be completely lost to git history...

obackhouse:
What's the deal with `add_tsymmetric_fragments`? It has a deprecation warning but as far as I can tell it's the intended function, were `make_tsymmetric_fragments` is the deprecated one?

cjcscott:
I think the preferred interface is currently to use `emb.symmetry.set_translations(supercell_mesh)` prior to adding any fragments, which will then automatically include all translationally related fragments.
This uses `qemb.create_transsym_fragments` rather than `add_tsymmetric_fragments` or `make_tsymmetric_fragments`, so I think we should depreciate both of those functions entirely.

I think I have an OK idea of how the different symmetries in Vayesta are being applied- translational symmetry is the only one using this approach (termed being applied in fragment context in the code comments, since it should apply to all included fragments) while rotation, inversion and mirror symmetries are handled via context managers within the fragmentation itself (termed as being within their own context).

However, as I think think there was much review done on these changes the examples and tests don't all reflect these updates- @basilib has caught some of them in #105  (eg. no longer supported approaches to rotational symmetries) but we should make a concerted effort to clean all this up here, as it is very unclear due to the number of old interfaces. If we then fix any tests broken by this, then merge this and fix any broken examples along with the rest in the above PR that seems  sensible to me? I've also mentally added updating the docs to have a proper discussion of this to my TODO list...

cjcscott:
I've updated this branch to remove conflicts with master; I'd suggest we merge this once tests pass, then I'll open a PR for clarifying the symmetries functions, as that's a suitably different set of things and can also involve updating the docs.

basilib:
I merged the example fixes into this branch, but there are still a few which require attention. I also wrote a quick script to run the examples.

basilib:
The interface for rotational symmetry has been changed to no longer accept letters (x,y,z) to specify the rotation axis and instead requires the use of an explicit vector. I have changed the examples to use explicit vectors but the tests use letters.

Is this an intended change or do we want to keep the option of specifying the axis by a letter?

vayesta/core/fragmentation/fragmentation.py
```diff
@@ -195,7 +195,7 @@ def rotational_symmetry(self, order, axis, center=(0,0,0), unit='Ang'):
         self.sym_register.append([])
         yield
         fragments = self.sym_register.pop()
-        fragments_sym = self.emb.create_rotsym_fragments(order, axis, center, fragments=fragments, unit=unit,
+        fragments_sym = self.emb.create_rotsym_fragments(order, tuple(axis), tuple(center), fragments=fragments, unit=unit,
```
cjcscott, line 198: This change is why `xyz` are no longer being accepted as valid axis options; they're being converted into tuples `("x",)` and no longer recognised later on. Why are we adding the tuple requirement here?

cjcscott: Fix which permits the use of lists to specify symmetry axes pushed.

maxnus:
> I think the preferred interface is currently to use `emb.symmetry.set_translations(supercell_mesh)` prior to initialising any fragmentations, which will then automatically include all translationally related fragments. This uses `qemb.create_transsym_fragments` rather than `add_tsymmetric_fragments` or `make_tsymmetric_fragments`, so I think we should depreciate both of those functions entirely.
> 
> I think I have an OK idea of how the different symmetries in Vayesta are being applied- translational symmetry is the only one using this approach (termed being applied in fragment context [here](https://github.com/BoothGroup/Vayesta/blob/c58d8eed8960f010c8be815f8b146273745ae837/vayesta/core/fragmentation/fragmentation.py#L69C12-L70), since it should apply to all included fragments) while rotation, inversion and mirror symmetries are handled via context managers within the fragmentation itself (termed as being within their own context).

Yes, this is correct 👍 All symmetry-related fragments are added via `qemb.create_symmetric_fragments` at the moment, which is invoked via creating fragments within a symmetry context manager

ghb24:
This is good to know, but also aware that once this PR is merged the discussions on this thread re. implementation of translational/other symmetry will be essentially lost! We need to make sure that this is clearly described in the docstrings / documentation / examples going forward...

cjcscott:
As I said above, I think at the same time as depreciating the alternative symmetry interfaces we should add a section into the quick start documentation discussing symmetry considerations. However, I'm a little wary that this PR already makes a lot of progress and delaying it for this purpose isn't actually beneficial, so would suggest merging this (once all examples are fixed) then I'll open a new PR updating symmetry stuff.

cjcscott:
I've updated the EDMET examples. Otherwise I've found:
-the tailoring examples appear to have had the same convergence problems on the commit they were introduced in (1e0c76a1c1a089238b5d0e8267150379ce871e31, see #68 as well...), so I think scrap them and start again. @abhishekkhedkar09 do you have any good scripts as an example of tailoring? I know you don't use delta tailoring so much, so maybe not for that.
-the population analysis functionality appears have been broken since the removal of ActiveSpace objects in favour of Cluster objects in 796e9a8e166f1a0ac56a6d6beaf55e048471111c. I would suggest fixing this probably falls outside the remit of this PR, but is definitely worthwhile; I don't think it'll be much effort(just requires adding functionality to add in the environmental 1rdm contribution) but is worth doing robustly. I'd suggest we leave this for now; I'll raise an issue to make sure it's documented, this might be a nice short PR for someone who wanted to have a go at it.

Just going through the last few examples now..

abhishekkhedkar09:
> I've updated the EDMET examples. Otherwise I've found: -the tailoring examples appear to have had the same convergence problems on the commit they were introduced in ([1e0c76a](https://github.com/BoothGroup/Vayesta/commit/1e0c76a1c1a089238b5d0e8267150379ce871e31), see #68 as well...), so I think scrap them and start again. @abhishekkhedkar09 do you have any good scripts as an example of tailoring? I know you don't use delta tailoring so much, so maybe not for that. -the population analysis functionality appears have been broken since the removal of ActiveSpace objects in favour of Cluster objects in [796e9a8](https://github.com/BoothGroup/Vayesta/commit/796e9a8e166f1a0ac56a6d6beaf55e048471111c). I would suggest fixing this probably falls outside the remit of this PR, but is definitely worthwhile; I don't think it'll be much effort(just requires adding functionality to add in the environmental 1rdm contribution) but is worth doing robustly. I'd suggest we leave this for now; I'll raise an issue to make sure it's documented, this might be a nice short PR for someone who wanted to have a go at it. -The ewf/other examples have a comment saying they were broken from from [3d82332](https://github.com/BoothGroup/Vayesta/commit/3d82332e8bd05f2aea4b2cb7bd9791c79dd640ac) (4th December 2021) so I've deleted them; I don't think it's worth the effort to fix them. I've also added rotational symmetry to `71-sc-h-ring.py` for computational efficiency.

alright..  I will add an illustrative example for Tailored CC.

obackhouse:
I think I am done - what's the status on the examples?

cjcscott:
I've just tweaked the delta tailoring example, and opened issues (#110 and #111) for the topics we've discussed here but not yet fixed. As such, I think this is ready to be merged.

obackhouse:
Ready to merge then

PR #96, opened 2023-05-25
Import cleaning

obackhouse:
Improves handling of optional dependencies including on installation, and will hopefully remove relative and star imports once I am done


obackhouse:
Also waiting on #92 so I can make sure `ebcc` is handled consistently

maxnus:
I would say relative imports are OK as long as they are explicit (with a leading `.`)
and a limited number of star imports are OK, if the imported module defines `__all__`.
What's the reason to get rid of them?

obackhouse:
PEP 8 recommends against relative imports: 

> Relative imports for intra-package imports are highly discouraged. Always use the absolute package path for all imports.

Absolute imports make it easier to avoid circular importing issues in complex packages.

As for star imports, PEP 8 does not explicitly mention them as far as I know, however in my opinion (and I think it's a common opinion) they're a bad idea. Not only do they pollute the namespace of the module, but they make it extremely difficult to chase down the origin of imported classes and methods, and generally reduce readability. They also prevent some static analysis and flaking tools on top of that.

I think in both cases, given that you can achieve the same behaviour with explicit and absolute imports, the latter is just the better option.

ghb24:
I certainly agree that following an execution thread of someone else's code is more difficult with star imports. But maybe thats because I'm still not using an IDE...

maxnus:
@obackhouse Where did you find this quote in PEP 8? Was it maybe a Python 2 version, before we had the dot-syntax?
https://peps.python.org/pep-0008/#imports says
> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

I don't have strong feelings about the star import, I agree they should generally be avoided.
Side note @ghb24 I now switched to PyCharm after so many years of vim it's the best thing ever :grin:  (just install ideavim to keep the commands)

cjcscott:
I think there is a decent benefit to consistency in our code, and I would err on the side of absolute imports just for clarity's sake. While relative imports are definitely sometimes easier to write, given the plethora of tools to allow automatic conversion between relative imports and absolute imports (including in PyCharm, which I'm glad you've finally come around to using!).

vayesta/core/types/wf/ccsd.py
obackhouse, line 1: @maxnus is this `vpyscf` import unnecessary now?

ghb24:
@obackhouse : Conflicts now with updated master...

obackhouse:
> @obackhouse : Conflicts now with updated master...

I'll sort it, don't think I'm done just yet anyway

obackhouse:
Good to merge now

PR #97, opened 2023-05-25
Update setup configuration for PEP 518

obackhouse:
Updates the setup configuration files so they're mostly contained in a `pyproject.toml`, with the compiled parts still handled in `setup.py`. Means fewer `.ini`, `.cfg`, `.rc` files floating around.

Wait on #92 and #96 (will probably fail tests for now)


obackhouse:
I don't think there's any way to get rid of `MANIFEST.in` if you're using `setuptools`, so that stays. `setup.py` still stays because it handles the compilation, but everything else is in `pyproject.toml`.

Do we need `CHANGELOG`? Tagged releases have their own changelog so it seems unnecessary.

cjcscott (approved):

PR #102, opened 2023-05-26
added level shift, diis space... options to CCSD

abhishekkhedkar09:
added level shift, diis space... options to CCSD. Also added fragment ID to the dummy pyscf object's output file name to avoid overwriting the meanfield pyscf output.


cjcscott:
Ah I totally missed the new CCSD options you added! I think we could definitely just add this into #92. Maybe we'd like to also add a test for this functionality, are there any simple systems which don't convergence without a level shift?

cjcscott:
Moved these changes into #92, as I just forgot to update for their addition in the rewrite.

PR #104, opened 2023-06-03
Adds bare run and 3.11 to CI

obackhouse:
The CI did not include a run that builds with none of the optional dependencies. This PR changes the standard runs to now use all optional dependencies, and adds a 'bare' run with none of them. It also adds a 3.11 run.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
     - cron: '0 2 * * *'
 
 jobs:
-  build:
+  full:
     name: python ${{ matrix.python-version }} on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


obackhouse:
FYI @cjcscott I think that `ebcc` wasn't actually getting imported in the CI on #92 :grimacing:

cjcscott:
I thought we already knew this? Hence lots of skipped tests? Good to run them though - 3.11 a good addition as well, let's see if any bugs gets turned up!

cjcscott (approved):

obackhouse:
Ah I may have misunderstood, did the `ebcc` tests pass locally? Should be fine if so.

obackhouse:
I've also edited this PR to drop CI support for python 3.7. It's EOL is in 3 weeks and `dyson` specifically requires >=3.8 anyway. I won't change the Vayesta requirements to disallow 3.7 for now, but I don't see much point in CI/CD for python versions past their EOL.

cjcscott:
Yeah, I've run tests with ebcc separately when plumbing in the interface etc

cjcscott:
I'll check what's happening with `cvxy` as a dependency here, the nicely skipping tests functionality appears to not be working..

obackhouse:
I fixed the `cvxpy` stuff, this can be merged

Issue #106, opened 2023-06-08
PBC error when using space_group_symmetry=True


cjcscott:
I've run into an issue when performing PBC calculations and enforcing `space_group_symmetry=True` where initialisation of an IAO fragmentation fails. Minimal example is
```
import pyscf.gto
import pyscf.lo
import pyscf.pbc
import numpy as np
import vayesta.ewf

cell = pyscf.pbc.gto.Cell()
cell.atom = ['He 0.0 0.0 0.0']
cell.a = 1.4 * np.eye(3)
cell.basis = 'def2-svp'
cell.output = 'pyscf.out'
cell.space_group_symmetry = True
cell.symmorphic = True
cell.build()

kpts = cell.make_kpts([2,2,2], space_group_symmetry=True, time_reversal_symmetry=True, symmorphic=True)

# Hartree-Fock with k-points
mf = pyscf.pbc.scf.KRHF(cell, kpts)
mf = mf.density_fit(auxbasis='def2-svp-ri')
mf.kernel()

emb = vayesta.ewf.EWF(mf)
with emb.iao_fragmentation() as f:
    f.add_atomic_fragment([0])

emb.kernel()
```
resulting in the error
```
          |  IAO:  computational basis= def2-svp  minimal reference basis= minao (automatically chosen)
Traceback (most recent call last):
  File "/users/k1819443/code/graphene/minex_vayesta_issue.txt", line 25, in <module>
    with emb.iao_fragmentation() as f:
  File "/scratch/users/k1819443/code/Vayesta/vayesta/core/qemb/qemb.py", line 1387, in iao_fragmentation
    return IAO_Fragmentation(self, minao=minao, **kwargs)
  File "/scratch/users/k1819443/code/Vayesta/vayesta/core/fragmentation/__init__.py", line 28, in IAO_Fragmentation
    return IAO_Fragmentation_RHF(emb, *args, **kwargs)
  File "/scratch/users/k1819443/code/Vayesta/vayesta/core/fragmentation/iao.py", line 43, in __init__
    self.refmol = pyscf.lo.iao.reference_mol(self.mol, minao=self.minao)
  File "/scratch/users/k1819443/miniconda3/envs/vayesta/lib/python3.10/site-packages/pyscf/lo/iao.py", line 135, in reference_mol
    pmol.build(False, False, basis=minao)
[...]

PR #107, opened 2023-06-09
Added informative error message when encountering pyscf space_group_symmetry bug.

cjcscott:
This relates to the errors mentioned in #106.

This returns a more informative error message when encountering a bug with generating IAOs in pyscf with space group symmetry enabled. This should only be present in a small number of pyscf versions, so requiring the user to update their input rather than trying to handle it internally seems a reasonable. The fix is to just disable space group symmetry for the mean-field calculation when initialising the IAO fragmentation; it can be turned back on afterwards if desired.

### DIFF
--- a/examples/ewf/solids/01-simple-sym.py
+++ b/examples/ewf/solids/01-simple-sym.py
@@ -6,15 +6,16 @@ import vayesta
 import vayesta.ewf
 
 
-# For an equivalent calculation enforcing symmetries, see examples/ewf/solids/01-simple-sym.py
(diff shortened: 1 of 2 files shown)


ghb24:
Looks good to me. Is it worth making this clear in the example too that uses space group/time-reversal symmetry?

cjcscott:
There isn't actually an example using space group symmetry currently- I'll add a duplicate of `01-simple.py` that enforces symmetries and includes instructions for avoiding this error.

PR #118, opened 2023-07-19
Variational embedding

abhishekkhedkar09:
an example and a couple of tests added, we could merge this now. Pygnme dependency remains an optional dependency. I leave it up to you @cjcscott whether we should uncomment the RDM1 computation. It is for sure not a limiting factor for cas > 8 orbitals calculations...


abhishekkhedkar09:
@cjcscott LGTM despite the codecov hit as it is caused by keeping pygnme out of CI and as an optional dependency...

cjcscott:
Yeah, that's grand- we've had more than a good go at getting pygnme to run in the CI. As I've said, really this isn't a Vayesta issue so much as a pygnme one. The codecov complaint isn't blocking (none of our tests are, technically) but good to know that it makes some noise for low coverage!

cjcscott (approved):

PR #119, opened 2023-07-28
cRPA Screening and external RPA corrections

cjcscott:
This adds support for cRPA screening of fragment interactions into Vayesta.
Various varieties of cRPA are possible. By default the conventional approach screening all interactions in the cluster via the screening equation is used; two phenomenological modifications are also supported (screening only the occupied-virtual interactions to avoid unintended reference modifications, and incorporating a term originating from the irreducible polarisability into the coupling of the cRPA to the cluster).

Thing left to do:
- [x] Print a warning if cRPA is used with any fragmentation other than CAS.
- [x] Add test for RPA external energy corrections.
- [x] Add screening example.


vayesta/core/qemb/fragment.py
```diff
@@ -1036,10 +1039,27 @@ def get_solver(self, solver=None):
...
+            hamil = hamil or self.hamil
+            cumulant = self.base.opts.ext_rpa_correction == "cumulant"
+            if self.base.opts.ext_rpa_correction not in ["erpa", "cumulant"]:
+                raise ValueError("Unknown external rpa correction %s specified.")
+            e_loc_rpa = hamil.calc_loc_erpa(*self._seris_ov[1:], cumulant)
+        return e_loc_rpa
+
     def get_frag_hamil(self):
+        if self.opts.screening is not None:
+            #if self
```
obackhouse, line 1054: is this commented statement needed?

vayesta/core/qemb/qemb.py
```diff
@@ -1561,6 +1589,7 @@ def _reset_fragments(self, *args, **kwargs):
     def _reset(self):
         self.e_corr = None
         self.converged = False
+        self.e_rpa = None
```
obackhouse, line 1592: is this ever initialised in `__init__`? If we definitely need it to be a variable in the object then make sure to initialise it rather than just assigning it later

vayesta/edmet/edmet.py
```diff
@@ -54,6 +54,17 @@ def eps(self):
         eps = eps.reshape(-1)
         return eps, eps
 
+    @property
+    def e_nonlocal(self):
+        try:
+            return self._e_nonlocal
+        except AttributeError:
+            return 0.0
```
obackhouse, line 62: as with the `e_rpa`, I think this pattern is bad practice - I think `self._e_nonlocal` should already be initialised as `None`

vayesta/core/qemb/fragment.py
```diff
@@ -1036,10 +1039,27 @@ def get_solver(self, solver=None):
...
+                                                            only_ov_screened=("ov" in self.opts.screening),
+                                                            log=self.log)
         # This detects based on fragment what kind of Hamiltonian is appropriate (restricted and/or EB).
         return ClusterHamiltonian(self, self.mf, self.log, screening=self.opts.screening,
-                                  cache_eris=self.opts.store_eris)
+                                  cache_eris=self.opts.store_eris, match_fock=self.opts.match_cluster_fock)
 
     def get_solver_options(self, *args, **kwargs):
-        raise AbstractMethodError
+        raise AbstractMethodError
```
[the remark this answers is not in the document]
cjcscott: will add back

vayesta/core/qemb/qemb.py
cjcscott: `__init__` actually calls `_reset`, so that's part of the function of adding it here.

vayesta/edmet/edmet.py
cjcscott: I've added the value into the `_reset` function for EDMET, following the pattern already used for `e_rpa` to initialise it there.

obackhouse: yeah I don't like that at all either :smile:

vayesta/tests/ewf/test_rpa_correction.py
```diff
@@ -0,0 +1,47 @@
...
+    """Tests with a complete bath in all clusters. This should give no nonlocal correction in any case."""
+    system = testsystems.water_631g_df
+
+    def test_rpa_correction(self):
+        enl, erpa = self.get_nl_energies("erpa", "full")
+        self.assertAlmostEqual(enl, 0.0)
+
+    def test_cumulant_correction(self):
+        enl, erpa = self.get_nl_energies("cumulant", "full")
+        self.assertAlmostEqual(enl, 0.0)
```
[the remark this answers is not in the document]
cjcscott: added

vayesta/edmet/edmet.py
obackhouse: oops that comment was meant for the above review - I think that all the variables should be initialised in `__init__` rather than `_reset` even if it duplicates that code, but I guess if the rest of vayesta follows the standard of not doing that then all good

cjcscott: Tbf it fitted pretty well with this one as well! Duplication could be justified here- but maybe having a `_initialise_variables` function just under `__init__` might be a reasonable compromise? In any case, I'd suggest this is for a potential future refactor rather than this PR.

obackhouse:
LGTM

cjcscott:
Use of mRPA with specifically graphene actually requires functionality in the next PR (use of non-PSD cderis with RIRPA). I have an example ready to go with that after this is merged.

abhishekkhedkar09 (approved):

PR #121, opened 2023-08-04
CCSDt' interface (and general EBCC interface updates)

cjcscott:
Updates the Vayesta interface for solvers from `ebcc`, adding support for CCSDt' active space solvers. Notable background updates are:

- Added functionality to apply rotations to CCSD wavefunctions and run solver calculations in a rotation of the cluster basis.
- Fixed factor of two conversion between certain spin components  of UCCSD T2 amplitudes from `ebcc`.
- Added tests and examples for using solvers from `ebcc` (CCSD, CCSDT and CCSDt').

### DIFF
--- a/vayesta/ewf/fragment.py
+++ b/vayesta/ewf/fragment.py
@@ -280,8 +280,8 @@ class Fragment(BaseFragment):
             self.log.debugv("Passing fragment option %s to solver.", attr)
             solver_opts[attr] = getattr(self.opts, attr)
 
-        has_actspace = ((solver == "TCCSD") or ("CCSDt" in solver) or
-                        (solver == "EBCC" and self.opts.solver_options['ansatz'] == "CCSDt"))
+        has_actspace = ((solver == "TCCSD") or ("CCSDt'" in solver) or
+                        (solver.upper() == "EBCC" and self.opts.solver_options['ansatz'] == "CCSDt'"))
         if has_actspace:
             # Set CAS orbitals
             if self.opts.c_cas_occ is None:
--- a/vayesta/ewf/fragment.py
+++ b/vayesta/ewf/fragment.py
@@ -101,7 +101,7 @@ class Fragment(BaseFragment):
         self.solver_results = None
 
     def set_cas(self, iaos=None, c_occ=None, c_vir=None, minao='auto', dmet_threshold=None):
-        """Set complete active space for tailored CCSD"""
+        """Set complete active space for tailored CCSD and active-space CC methods."""
         if dmet_threshold is None:
             dmet_threshold = 2*self.opts.bath_options['dmet_threshold']
         if iaos is not None:
--- a/vayesta/ewf/fragment.py
+++ b/vayesta/ewf/fragment.py
@@ -100,6 +100,11 @@ class Fragment(BaseFragment):
         # For self-consistent mode
         self.solver_results = None
 
+    def _reset(self, *args, **kwargs):
+        super()._reset(*args, **kwargs)
+        # Need to unset these so can be regenerated each iteration.
+        self.opts.c_cas_occ = self.opts.c_cas_vir = None
+
     def set_cas(self, iaos=None, c_occ=None, c_vir=None, minao='auto', dmet_threshold=None):
         """Set complete active space for tailored CCSD and active-space CC methods."""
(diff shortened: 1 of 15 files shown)


vayesta/ewf/fragment.py
```diff
@@ -280,8 +280,9 @@ def get_solver_options(self, solver):
             self.log.debugv("Passing fragment option %s to solver.", attr)
             solver_opts[attr] = getattr(self.opts, attr)
 
-        if solver.upper() == 'TCCSD':
-            solver_opts['tcc'] = True
+        has_actspace = ((solver == "TCCSD") or ("CCSDt" in solver) or
+                        (solver == "EBCC" and self.opts.solver_options['ansatz'] == "CCSDt"))
```
obackhouse, line 284: Can we maybe do `solver.upper() == "EBCC"`? I stylise `ebcc` as lower-case

obackhouse: as in, I stylise the code that way, but the actual solver classes in `ebcc` are indeed caps, so I guess maybe this makes sense

vayesta/tests/solver/test_ebcc.py
```diff
@@ -0,0 +1,105 @@
...
+# Note that ebCC currently doesn't support density fitting, so we're just testing non-DF results here.
+
+
+class TestEBCC(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        try:
+            import ebcc
+        except ImportError:
+            pytest.skip("Requires cvxpy")
```
obackhouse, line 21: Typo

vayesta/tests/solver/test_ebcc.py
```diff
@@ -0,0 +1,105 @@
...
+        return self._test('water_cation_sto3g', 'uhf', 'CCSDT')
+
+
+class TestEBCCActSpace(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        try:
+            import ebcc
+        except ImportError:
+            pytest.skip("Requires cvxpy")
```
obackhouse, line 68: and again

vayesta/ewf/fragment.py
cjcscott: It depends- I was previously just using `solver.upper()` in the solver selection which would have done this automatically, but to support CCSDt' as an option I changed that in this PR. I was wondering about your prefered stylising (I plumped for `ebCC` above)- I could just change to `solver.upper()`for this specific check?

cjcscott: I've also just realised this is checking `=="CCSDt"` not `=="CCSDt'"`, which I'll fix at the same time. That prime is a little awkward, but I'm assuming CCSDt is a separate thing? It's been a while since I checked the literature for this.

cjcscott: That's fair, I don't think I've used anywhere else and have updated the above to reflect that.
I've updated this so that capitalisation of `eb` is arbitrary, along with `cc` if no ansatz is specified in the solver. To be compatible with `ebcc`'s ansatz specification we need the ansatz to be case-sensitive, so for instance we'd allow `ebcc`, `ebCCSD`, or `EBCCSD` but not `ebccSD`. Does that sound like a reasonable compromise?

obackhouse: sure, any case can work really if we just throw in an `upper`

cjcscott: Given `ebcc` requires case-sensitivity in ansatz specification I think requiring it for Vayesta is acceptable- just so that we don't end up with unexpected interactions down the line eg. someone doesn't think they're running a CCSDt calculation down the line and ending up with an untenable CCSDT? Basically, given that capitalisation has meaning in CC ansatzes we probably can't just always use `solver.upper()` without a bit more documentation of supported approaches.

obackhouse: so I mean for `EBCC` specifically, like `ebcc == ebCC == EBCC`, but yea if you're specifying a specific CC ansatz then case sensitivity is important.

cjcscott: Sweet, yeah that's what we've got

abhishekkhedkar09 (dismissed):

abhishekkhedkar09 (approved):

PR #122, opened 2023-08-08
RIRPA Optimisations and Bath Construction

cjcscott:
This PR adds various updates to the RIRPA functionality within Vayesta. The three major components of this are:

- Added support for RIRPA calculations to be performed with negative CDERI components.
- Added new class specialised to performing efficient restricted, dRPA calculations. This is implemented using the existing numerical integration framework, and should correspond to a factor of ~2 speedup to the limiting steps of the computation.
- Added new bath construction based on the RPA dd moment, as used in the mRPA paper.

These all have various tests and examples added. I also fixed a small bug in the optimisation of the quadrature spacing, where the search bounds were set unnecessarily tightly in some situations, which may result in an improved grid compared to previously. For practical calculations the difference in final results is tiny though (eg. as used in mRPA tests).

Things left to do:
- [ ] Add RIRPA test on solids/graphene.


ghb24:
Is the example for a single k-point? It might be worth putting that explicitly, just to make it clear to users how to extend this further for reproduction of paper results.

ghb24:
Also, I know that this PR may not be the right place, but should we have an example of cRPA screening with active space 'fragments', just to show off this functionality?

cjcscott:
I had gone for a single kpoint for the sake of keeping the examples fast, but I've changed it now to instead run with a minimal basis and (221) kpoints. I've also added an example running a CAS calculation on benzene with bare and screened interactions, just to show this off!

PR #126, opened 2023-08-16
ebcc wavefunctions

cjcscott:
This adds CC wavefunction objects which use `ebcc` as a backend, rather than `pyscf`, along with tests for this functionality.  With this we'll have a bit more flexibility in our wavefunction specifications, as it's compatible with arbitrary CC ansatzes. This is a prerequisite to making more use of coupled electron-boson clusters.

I've also included a small tweak to the `pyproject.toml` which allows tests tagged veryslow to be run; previously our configuration was actually blocking running these tests, even when requested.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ dyson = [
     "dyson @ git+https://github.com/BoothGroup/dyson@master",
 ]
 ebcc = [
-    "ebcc"
+    "ebcc @ git+https://github.com/BoothGroup/ebcc@master",
 ]
 
 pygnme = [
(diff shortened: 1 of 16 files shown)


cjcscott:
That sounds good- I thought that might be the case, but wanted to verify that was the only cause of failures before bothering you with it

obackhouse:
done, will be live once the CI runs through

pyproject.toml
```diff
@@ -110,7 +110,7 @@ exclude_lines = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "--import-mode=importlib -k 'not veryslow'"
+addopts = "--import-mode=importlib"
```
obackhouse, line 113: I don't think this should be the default here - this means that when people run `pytest vayesta` the `veryslow` tests will run, and it takes ages. I think this should just be controlled in the CI if we want them to always run, or just in the nightly scheduled run?

ghb24 (approved):

PR #127, opened 2023-08-17
Bugfix PR: This fixes a bug introduced while re-enabling calculation of rdm1 in variational embedding

abhishekkhedkar09:


abhishekkhedkar09:
so basically the codecov patch shows an error as the tests are skipped. The actual tests have covered the said lines, I checked!

PR #128, opened 2023-08-23
Boson bath construction

cjcscott:
This adds a general interface for the construction and use of bosonic baths within embedding calculations, along with support for these new options within EWF. EDMET isn't yet rewritten to make use of this new interface, but will be in future and the approach used there remains supported. Examples and tests for the new functionality are included.


vayesta/core/bosonic_bath/projected_interactions.py
```diff
@@ -0,0 +1,238 @@
...
+        return (poa, pob), (pva, pvb)
+
+    def _loop_df(self, blksize=None):
+        nao = self.mf.mol.nao
+        df = self.mf.with_df
+        naux = self.naux
+        if blksize is None:
+            blksize = int(1e9 / naux * nao * nao * 8)
+        # PBC:
+        if hasattr(df, 'sr_loop'):
```
obackhouse, line 217: Is this always just grabbing the integrals at the gamma point? `pbc.df.DF` has `loop` implemented to do this, so that gamma point DF objects are compatible with molecular post-HF methods

vayesta/tests/core/bath/test_bosonic_bath.py
```diff
@@ -0,0 +1,50 @@
...
+        occ = self._get_occupation("full", None)
+        values = np.array([np.amin(occ), np.amax(occ), np.mean(occ)])
+        known_values = np.array([5.23868995e-17, 3.12442358e-02, 5.99949472e-04])
+        self.assertAllclose(values, known_values)
+
+    def test_target_dmet_no_project(self):
+        occ = self._get_occupation("dmet", None)
+        values = np.array([np.amin(occ), np.amax(occ), np.mean(occ)])
+        known_values = np.array([1.62184716e-08, 9.82891292e-03, 5.32748449e-04])
+        self.assertAllclose(values, known_values)
```
obackhouse, line 50: Missing newline

vayesta/core/bosonic_bath/projected_interactions.py
cjcscott: Yeah, I think this is just grabbing the gamma point integrals; I just extracted it from `get_cderis` in `core/eris.py`. I'm assuming we could similarly refactor that function then?

cjcscott: Looking at that code the difference is that this also treats non positive definite eris (ie. 2D pbc systems); while it's very similar functionality I think we need to be able to treat that case, so sadly can't make that simplification for now.

ghb24:
Saw that there was a comment from Ollie about multiple ways to get the eris when needed. If it is not appropriate to use a particular interface due to it's lack of support for some functionality, can you please leave a comment in the code at the appropriate place with a TODO to harmonise the interfaces in the future with the required functionality. This should make it easier to understand the limitations for future developers.

cjcscott:
This is a fundamental constraint on the pyscf cderi routines for pbc vs molecular systems, rather than an interface limitation on our side to be fixed/harmonised later. The pbc routines have to return sign information that the molecular ones don't- we could wrap this and generalise, but there aren't major gains to code reuse or similar to be had here.

PR #130, opened 2023-09-06
black formatting the latest change to file ccsd.py

abhishekkhedkar09:


cjcscott (approved):
Looks good (assuming tests pass with modified pyscf specification)!

cjcscott:
This failure is just because codecov apparently sets a default patch coverage target of the current project coverage; in this case it's not a problem, but if anyone want to fix this I would expect explicitly setting a patch coverage target to override it.

PR #133, opened 2023-09-20
Fixes #132

### LINKED ISSUE #132: Error with Brueckner SCMF: MO-coefficients are not orthonormal
/The orthogonalization of `mo_coeff` fails [here](https://github.com/BoothGroup/Vayesta/blame/56248bb89c098c667b61de4dfe891dda5a73b56f/vayesta/core/scmf/brueckner.py#L59) if linear dependencies are removed in calculations with Brueckner SCMF. This is when `mf.mol.nao_nr() - mf.mo_coeff.shape[1] > 0`.  We should have a more robust orthogonalization and perhaps need to use the updated overlap matrix to alleviate the error. I am just posting this here so that this problem is pinned. 
The MWE that fails is as follows. The overlap matrix in the following is ill-conditioned and in update_mo_coeff we calculate its inverse [here](https://github.com/BoothGroup/Vayesta/blame/56248bb89c098c667b61de4dfe891dda5a73b56f/vayesta/core/scmf/brueckner.py#L63) 

```
import vayesta, pyscf
from pyscf.gto import Mole
from vayesta.ewf import EWF
from pyscf.scf import RHF
from vayesta.misc import molecules
import numpy as np
basis, r, proj, lenfrag, bruecknerSCMF, mode = "6-31g", 1.2, 1, 2, True, 'external'

mol = Mole()
mol.atom = vayesta.misc.molecules.chain('H', 20, r)
mol.basis = f'{basis}'
mol.unit = 'B'
mol.build()
mf = RHF(mol)
s = mol.intor('cint1e_ovlp_sph')
if np.linalg.eigh(s)[0][0] < 1e-7:
    print(f"CAUTION: for r={r}, a linear dependency was removed using the pyscf addon ={np.linalg.eigh(s)[0][:8]}")
    mf = pyscf.scf.addons.remove_linear_dep(mf,threshold=1e-07, lindep=1e-7,
        cholesky_threshold=1e-10, force_pivoted_cholesky=True)


mf = mf.density_fit()
mf.kernel()
print(mol.nao_nr() - mf.mo_coeff.shape[1])
assert mf.converged

emb = EWF(mf)
fci_frags = []
natm = mf.mol.natm
with emb.iao_fragmentation() as f:
    for i in range(0, natm, lenfrag):
        fci_frags.append(f.add_atomic_fragment(list(range(i ,i + lenfrag)),
                solver='FCI', bath_options=dict(bathtype='dmet'),auxiliary=True) )
    ccsd_frag = f.add_full_system(solver='extCCSD',bath_options=dict(bathtype='full'))
ccsd_frag.add_external_corrections(fci_frags, correction_type=mode, projectors=proj)
if bruecknerSCMF:
    emb.brueckner_scmf()
emb.kernel()
```

maxnus:
Addressing #132, but was not able to run the attached example to verify

### DIFF
--- a/vayesta/core/scmf/brueckner.py
+++ b/vayesta/core/scmf/brueckner.py
@@ -60,9 +60,11 @@ class Brueckner_RHF(SCMF):
             bmo_occ = np.dot(self._mo_orig, v)[:, -nocc:]
 
         # Virtual space
-        dm_vir = np.linalg.inv(ovlp) - np.dot(bmo_occ, bmo_occ.T)
-        e, v = scipy.linalg.eigh(dm_vir, b=ovlp, type=2)
-        bmo_vir = v[:, -nvir:]
+        r = dot(mf.mo_coeff.T, ovlp, bmo_occ)
+        e, v = np.linalg.eigh(np.dot(r, r.T))
+        assert np.allclose(e[:nvir], 0)
+        assert np.allclose(e[nvir:], 1)
+        bmo_vir = np.dot(mf.mo_coeff, v)[:, :nvir]
 
         assert (bmo_occ.shape[-1] == nocc) and (bmo_vir.shape[-1] == nvir)
         mo_coeff = np.hstack((bmo_occ, bmo_vir))


abhishekkhedkar09:
> Addressing #132, but was not able to run the attached example to verify

I am unsure why you saw the error... I tested it for the same and works fine. This fix alleviates the orthogonality error. LGTM!

PR #134, opened 2023-09-20
Refactors SCMF UHF code

maxnus:
This is to reduce code duplication in the UHF versions of SCMF, which now calls the RHF update twice.

Together with #133 this addresses #132 for the UHF case


abhishekkhedkar09:
I tested, and both #133 and this one fixed the issue and passed the regression test. we can merge!

PR #142, opened 2023-09-23
Drops support for Python 3.7

maxnus:
I believe we don't support Python 3.7 any longer?

edit: we also spent a lot of time building PySCF, let's cache it


obackhouse:
thanks, don't know why I didn't do this when I removed 3.7 from the CI

obackhouse (approved):

PR #144, opened 2023-09-24
Change PySCF version

maxnus:
I'm not seeing [PySCF issue 1846](https://github.com/pyscf/pyscf/issues/1846) on my side, may have been fixed?

edit: Nevermind, ran with an older build by accident. I made a PR to fix here: https://github.com/pyscf/pyscf/pull/1876
In the meantime, should we use v2.3.0 and stick to PyPI releases from now on?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
     "scipy>=1.1.0,<=1.10.0",
     "h5py>=2.7",
     "cvxpy>=1.1",
-    "pyscf @ git+https://github.com/pyscf/pyscf@master",
+    "pyscf==2.3.0",
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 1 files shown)


obackhouse:
> In the meantime, should we use v2.3.0 and stick to PyPI releases from now on?

I was a strong supporter of this - ultimately as a group it was decided that we want to be current with PySCF's master branch in order to support new PySCF features without having to wait on their slow release schedule (not that our release schedule is any better, we should really work on that)

The main reason I support it is because it makes building wheels trivial, but without relying on PySCF wheels then building our wheels is very much not trivial because we'd have to package PySCF ourselves too...

maxnus:
>I was a strong supporter of this - ultimately as a group it was decided that we want to be current with PySCF's master branch in order to support new PySCF features without having to wait on their slow release schedule (not that our release schedule is any better, we should really work on that)
>
>The main reason I support it is because it makes building wheels trivial, but without relying on PySCF wheels then building our wheels is very much not trivial because we'd have to package PySCF ourselves too...

In my opinion, there is literally no downside in having *a* branch sticking to PySCF releases so that we can publish to PyPI
from there. You can still build from source, just as you can now.

I guess the only question is do we want to do this with `main`... in my opinion yes, but we can also create a release branch.

PR #148, opened 2023-11-06
Some changes to facilitate newly implemented CCSDt

abhishekkhedkar09:
Added Active space support for EBCC CCSDt; Avoid considering auxiliary for orthogonality tests for EC-CC or TCC calculations. Changes in hamiltonian.py to return a np array of (2, modim, modim) instead of a tuple, the tuple results in errors in ebcc for UHF. PYSCF-wide default is the np array.

### DIFF
--- a/vayesta/core/qemb/qemb.py
+++ b/vayesta/core/qemb/qemb.py
@@ -1574,7 +1574,7 @@ class Embedding:
         for s in range(nspin):
             nmo_s = tspin(self.nmo, s)
             nelec_s = tspin(nelec, s)
-            fragments = self.get_fragments(contributes=True, active=True, flags=dict(is_secfrag=False))
+            fragments = self.get_fragments(contributes=True, flags=dict(is_secfrag=False))
             if not fragments:
                 return False
             c_frags = np.hstack([tspin(x.c_frag, s) for x in fragments])
--- a/vayesta/core/qemb/qemb.py
+++ b/vayesta/core/qemb/qemb.py
@@ -1574,7 +1574,7 @@ class Embedding:
         for s in range(nspin):
             nmo_s = tspin(self.nmo, s)
             nelec_s = tspin(nelec, s)
-            fragments = self.get_fragments(active=True, flags=dict(is_secfrag=False))
+            fragments = self.get_fragments(contributes=True, active=True, flags=dict(is_secfrag=False))
             if not fragments:
                 return False
             c_frags = np.hstack([tspin(x.c_frag, s) for x in fragments])
--- a/vayesta/ewf/fragment.py
+++ b/vayesta/ewf/fragment.py
@@ -306,6 +306,7 @@ class Fragment(BaseFragment):
         has_actspace = (
             (solver == "TCCSD")
             or ("CCSDt'" in solver)
+            or ("CCSDt" in solver)
             or (solver.upper() == "EBCC" and self.opts.solver_options["ansatz"] == "CCSDt'")
         )
         if has_actspace:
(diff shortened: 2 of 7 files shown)


abhishekkhedkar09:
> Is it better to fix
> 
> > Changes in hamiltonian.py to return a np array of (2, modim, modim) instead of a tuple, the tuple results in errors in ebcc for UHF
> 
> in `ebcc` instead? I think this line would just need changing:
> 
> https://github.com/BoothGroup/ebcc/blob/ff5240121b23e43c2e760968946f51900b13c805/ebcc/uebcc.py#L448
> 
> to `np.asarray(mf.get_fock())`? or is the error somewhere else?

That's what I thought at first. But I tested what pyscf returns for UHF mf.get_fock() and it is indeed (2, nmo, nmo ) and not a tuple. I recall this was put in as pad_to_match fix for active spaces with unequal numbers of electrons/orbitals in alpha and beta channels.

obackhouse:
cool, up to you

vayesta/core/qemb/qemb.py
```diff
@@ -1574,7 +1574,7 @@ def _check_fragmentation(self, complete_occupied=True, complete_virtual=True, to
         for s in range(nspin):
             nmo_s = tspin(self.nmo, s)
             nelec_s = tspin(nelec, s)
-            fragments = self.get_fragments(active=True, flags=dict(is_secfrag=False))
+            fragments = self.get_fragments(contributes=True, active=True, flags=dict(is_secfrag=False))
```
maxnus, line 1577: `contributes` implies `active`

vayesta/ewf/fragment.py
```diff
@@ -306,7 +306,9 @@ def get_solver_options(self, solver):
         has_actspace = (
             (solver == "TCCSD")
             or ("CCSDt'" in solver)
+            or ("CCSDt" in solver)
```
maxnus, line 309: `"CCSDt'" in solver` implies `"CCSDt" in solver` (acceptable if you want to be very explicit though)

vayesta/ewf/fragment.py
```diff
@@ -306,7 +306,9 @@ def get_solver_options(self, solver):
...
             or (solver.upper() == "EBCC" and self.opts.solver_options["ansatz"] == "CCSDt'")
+            or (solver.upper() == "EBCC" and self.opts.solver_options["ansatz"] == "CCSDt")
```
maxnus, line 311: better one line,  `...self.opts.solver_options["ansatz"] in {"CCSDt", "CCSDt'"}`

vayesta/solver/hamiltonian.py
```diff
@@ -639,7 +640,7 @@ def get_fock(self, with_vext=True, use_seris=True, with_exxdiv=False):
                 + einsum("iipq->pq", gab[oa, oa])  # Coulomb
                 - einsum("ipqi->pq", gbb[ob, :, :, ob])
             )  # Exchange
-            fock = ((fock[0] + dfa), (fock[1] + dfb))
+            fock = np.asarray(((fock[0] + dfa), (fock[1] + dfb)))
```
maxnus, line 643: I don't think this can be packed into a single array, since `fock[0]` and `fock[1]` can have different shapes?

vayesta/core/qemb/qemb.py
abhishekkhedkar09: true! fixed.

vayesta/solver/hamiltonian.py
abhishekkhedkar09: It can not be if the shapes are different. Wouldn't Fock matrices for alpha and beta always have the same shape?

PR #151, opened 2023-11-08
Adds exception for restriced open-shell mean-field

maxnus:
As title


basilib:
I've added an automatic conversion to UHF with a warning to deal with this.

PR #166, opened 2024-04-23
Allow calculating the democratically parititoned energy when using MPI

basilib:


ghb24 (commented):
Looks good to merge after this comment addressed.

PR #171, opened 2024-07-31
Add support for directories in cluster mol output

maxbortone:
The definition of the output on the cluster `mol` object didn't support setting it to a particular path. This meant that the output file from the cluster `mol` object was always stored in the current working directory, rather than where the output file of the main mean-field `mol` object was set to. This pull request fixes it, so that all PySCF output files are now stored in the same user defined directory.


basilib:
Looks good!

basilib:
Thank you for fixing this

PR #173, opened 2024-08-30
Adds local CCSD(T) energy

basilib:
Adds local CCSD(T) correction supporting UHF, RHF, molecules and solids.


ghb24:
Can we get a test + example for this PR?

Issue #182, opened 2024-11-30
Incompatible with `ebcc==1.6.0`


obackhouse:
Vayesta will be incompatible with `ebcc==1.6.0`, shouldn't be too hard to fix

basilib:
This issue has been resolved in #186  and #187.

PR #183, opened 2024-11-30
Freeze `ebcc` to 1.5.0

obackhouse:
Freezes the `ebcc` version until #182 is addressed


basilib (commented):
Looks good. Hopefully we can make a more complete fix to maintain compatibility soon.

